PR #110, opened 2015-01-12
Fix deprecation warnings on julia 0.4

timholy:
I hope I fixed these in a way that doesn't break compatibility with 0.3. Hopefully Travis will let us know.


timholy:
Hmm, I must have been working off an old version; it seems like much of this has been fixed on master. I'll clean it up with just the new stuff.

timholy:
Looks like you might need to flip the switch on this over at Travis, BTW. Visit https://travis-ci.org/profile/ihnorton and then hit the toggle for this repo.

timholy:
OK, this should be it.

I also only just noticed there aren't tests in the `tests

PR #112, opened 2015-01-13
Turn on tests in runtests.jl, and fix expr comparison in test/wrap_c

timholy:
This also adds `Pkg.build("Clang")` to the tests.

One more issue, though: I'm not sure if the version of julia on travis was built with `BUILD_LLVM_CLANG=1`.


timholy:
Bummer. Hopefully this should be helpful, though, for local tests with `Pkg.test("Clang")`.

If you want to run tests on travis, I think there might be two options:
- Ask Elliot to include Clang in his ppas
- Build julia first, and then test Clang.jl

The likely problem with the 2nd approach is timeouts (50min max).

PR #113, opened 2015-01-13
Prevent overwriting structs with typedefs

timholy:
This addresses some problems I encountered in wrapping CUDA6.5. I was finding that a struct definition, followed by a typedef with the same name, would result in the trivial `typealias dim3 dim3` being written to the common.jl file.


ihnorton:
Please merge at will.

I'll work on figuring out how to get CI going so that this can be worked
into a more maintainable state.

PR #119, opened 2015-02-16
Fix some issues

rollingthunder:
I am trying to create a wrapper from some C headers and came across these issues.

Hope this helps.

Thanks for a very useful package.


ihnorton:
:+1:  thanks.

PR #121, opened 2015-03-17
I can haz continuous integration?

ihnorton:
FYI: also bumps REQUIRE to 0.4. I hope this won't be a problem for package authors as long as the output remains backward-compatible with 0.3. Moving to 0.4 will let me rewrite the libclang interface to use byval structs instead of the nasty memcpy hack. This should improve maintainability, and get us any improvements made to libclang since llvm 3.1 (once the interface is self-regenerating).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
     #- JULIAVERSION="juliareleases"
     - JULIAVERSION="julianightlies"
 before_install:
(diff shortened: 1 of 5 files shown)


kmsquire:
Probably obvious, but just in case: because of the 0.4 dependency, when you
tag this, can you be sure to bump the minor version number, and not just
the patch?

Cheers!
   Kevin

ihnorton:
Will do.

PR #123, opened 2015-04-03
Generate compatible Int*/UInt* conversions

kmsquire:
In particular, this changes
```julia
const AV_PIX_FMT_VDPAU_MPEG1 = int32(37)
```
into
```julia
const AV_PIX_FMT_VDPAU_MPEG1 = @compat (Int32)(37)
```

(The parenthesis are somewhat annoying, but mostly harmless.)

### DIFF
--- a/src/wrap_c.jl
+++ b/src/wrap_c.jl
@@ -180,16 +180,20 @@ cl_to_jl = @compat Dict{Any,Any}(
     )
 
 int_conversion = @compat Dict{Any,Any}(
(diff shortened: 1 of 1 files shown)


kmsquire:
Not sure if this should be an option, or done by default. The benefits of making it an option is that, eventually, users might want to stop worrying about v0.3 compatibility.  But then, Clang.jl might be an entirely different beast by then, etc.

ihnorton:
LGTM

PR #124, opened 2015-04-13
Use C_NULL instead of zero(Ptr{T}) in Array_x_Ptr constructors

kmsquire:


Gnimuc:
It's easy to build a customed wrapper generator for this kind of non-general use case, specifically, auto-generating these constructors by overloading `wrap!(ctx::AbstractContext, cursor::CLStructDecl)`.

PR #128, opened 2015-05-18
support UTF8 headers paths

vtjnash:
joinpath may return a utf8 string in many cases, it's nice to allow it to be supported here


ihnorton:
feel free to merge/push (pretty sure you have commit bit, let me know if
not)

vtjnash:
i wanted to ensure this actually worked first, since i made the edit directly in the github gui. but travis seems happy with it, so i guess it's good?

ihnorton:
LLVM supports UTF8, should be ok.

PR #130, opened 2015-05-24
Use relative path and `@__FILE__` instead of `Pkg.dir`

yuyichao:
In order to support installing outside `.julia`. Either for system wise install or for developing outside `.julia`.

Also includes a few whitespace fixes.


ihnorton:
Thanks!

PR #137, opened 2015-09-18
Proposed - Fixes for v0.4 compat

mdcfrancis:
Note - I had to comment out UInt128 from the exports

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,3 +1,10 @@
+if VERSION >= v"0.4.0-dev+6521"
+    using Base.Libdl
+    dlext = Base.Libdl.dlext
+else
+    dlext = Base.Sys.shlib_ext
+end
+
 find_llvm() = begin
     if haskey(ENV, "TRAVIS")
         return readchomp(`which llvm-config-3.3`)
@@ -22,4 +29,4 @@ if (!ispath("../usr/lib"))
     run(`mkdir ../usr/lib`)
 end
 
-run(`mv libwrapclang.$(Libdl.dlext) ../usr/lib`)
+run(`mv libwrapclang.$(dlext) ../usr/lib`)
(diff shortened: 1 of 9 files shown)


ihnorton:
Thanks, could you rebase against master? Not sure where else the conflict warning could be coming from.

mdcfrancis:
Looks better now

ihnorton:
Thanks!

deps/ext.jl
```diff
@@ -1,10 +1,13 @@
...
                             "usr", "lib" , "libwrapclang.so")
 else
-    const libwci = Libdl.find_library(["libwrapclang",],
-                                      [joinpath(dirname(@__FILE__),
-                                                "usr", "lib")])
+  if VERSION > v"0.4-"
+    const libwci = Base.Libdl.find_library(["libwrapclang",],[Pkg.dir("Clang", "deps", "usr", "lib"),])
+  else
+    const libwci = find_library(["libwrapclang",],[Pkg.dir("Clang", "deps", "usr", "lib"),])
+  end
```
yuyichao: WTH????
Why is it necessary to change back to `Pkg.dir` breaks `LOAD_PATH`?

deps/build.jl
```diff
@@ -22,4 +29,4 @@ if (!ispath("../usr/lib"))
     run(`mkdir ../usr/lib`)
 end
 
-run(`mv libwrapclang.$(Libdl.dlext) ../usr/lib`)
+run(`mv libwrapclang.$(dlext) ../usr/lib`)
```
yuyichao: Also I don't see the point of this change either.
`Clang.jl` requires julia 0.4 and AFAICT `Libdl.dlext` is available ever since `Libdl` is introduced. It's not like you are removing all direct references to `Libdl` (not even in this file) either.

Also AFAICT, the version that introduces Libdl is `0.4.0-dev+3840` (Compat.jl uses `0.4.0-dev+3844`) and not the version you specify here.

PR #139, opened 2015-10-02
Fix deprecation warnings and breakage caused by #137

yuyichao:
* Revert 2 useless changes in #137 
* `Union()` -> `Union{}`


ihnorton:
Thanks. Sorry about that.

PR #153, opened 2016-01-23
RFC: Generate safer wrapper function

yuyichao:
* Remove type constraint in the wrapper so that it is compatible with the
  `ccall` GC root mechanism.
* Map `char*` to `Cstring` and `wchar_t*` to `Cwstring`

Closes https://github.com/ihnorton/Clang.jl/issues/152


yuyichao:
Not sure where's the best place to mention the necessary user-supplied methods.

vchuravy:
@ihnorton Is there anything specific holding this PR up?

PR #154, opened 2016-02-05
Nicer build error if clang headers not found

jiahao:
If `make` of the build wrapper fails during `Pkg.build`, checks if
`BUILD_LLVM_CLANG=1` was specified in `$JULIA_HOME/Make.user`. If not, print an
error message hinting to the user that they should specify it and try again.


ihnorton:
Thanks!

PR #155, opened 2016-02-15
Enable Wrapping of Petsc

JaredCrean2:
This PR makes the changes necessary to wrap Petsc as needed for PETSc.jl.  The only change is that the buffer of common items (typedefs, enums etc.) gets read by the rewriter before the function definitions.  This enables the rewriter to change the function definitions to reflect any changes made to the common items.


ihnorton:
Is this still necessary when `petscconf.h` is available?

JaredCrean2:
Yes, because changes that are made to typedefs can only be made to the typedef declaration and then the usage in function declarations changed.  So the rewriter has to see the declarations of the common items before the function declarations

PR #156, opened 2016-03-07
Optimistically create unions.

vchuravy:
Without this Clang.jl fails to generate code for the `VkClearValue` as defined below. The problem is that both member sizes are unkown and thus equal to zero leading to an IndexOutOfBound error.

``` C
#include <stdint.h>

typedef union VkClearColorValue {
    float       float32[4];
    int32_t     int32[4];
    uint32_t    uint32[4];
} VkClearColorValue;

typedef struct VkClearDepthStencilValue {
    float       depth;
    uint32_t    stencil;
} VkClearDepthStencilValue;

typedef union VkClearValue {
    VkClearColorValue           color;
    VkClearDepthStencilValue    depthStencil;
} VkClearValue;
```

### DIFF
--- a/src/wrap_c.jl
+++ b/src/wrap_c.jl
@@ -322,6 +322,11 @@ function largestfield(cu::UnionDecl)
             maxelem = i
         end
     end
+    if maxelem == 0
+        warn("  all field sizes are unknown.")
+        maxelem = 1
+    end
+
     fields[maxelem]
 end
 


ihnorton:
Hmm, it should be able to get the sizes of those member structs. Are you sure all of the headers are being found correctly?

With this change, how would you use the result? Clangjl doesn't generate Julia Unions (which are not like C unions), it just generates a type that will match the size of the C union for ABI compatibility purposes. You would have to do some casting / `convert` on top of that to make it useful.

vchuravy:
I was contemplating creating a bitstype that has the correct size for ABI purposes and then adding `cconvert`s from the original types to the bitstype, but I hadn't had time to work on it. 

The idea came from reading: http://llvm.lyngvig.org/Articles/Mapping-High-Level-Constructs-to-LLVM-IR#27

The code generated right now is (with this patch is):

```julia
# Automatically generated using Clang.jl wrap_c, version 0.0.0

using Compat

immutable Array_4_Cfloat
    d1::Cfloat
    d2::Cfloat
    d3::Cfloat
    d4::Cfloat
end

zero(::Type{Array_4_Cfloat}) = begin  # /home/wallnuss/.julia/v0.4/Clang/src/wrap_c.jl, line 268:
        Array_4_Cfloat(fill(zero(Cfloat),4)...)
    end

immutable VkClearColorValue
    _VkClearColorValue::Array_4_Cfloat
end

immutable VkClearDepthStencilValue
    depth::Cfloat
    stencil::UInt32
end

immutable VkClearValue
    _VkClearValue::VkClearColorValue
end
```

vchuravy:
And on master (after removing the union)

```julia
# Automatically generated using Clang.jl wrap_c, version 0.0.0

using Compat

immutable Array_4_Cfloat
    d1::Cfloat
    d2::Cfloat
    d3::Cfloat
    d4::Cfloat
end

zero(::Type{Array_4_Cfloat}) = begin  # /home/wallnuss/.julia/v0.4/Clang/src/wrap_c.jl, line 266:
        Array_4_Cfloat(fill(zero(Cfloat),4)...)
    end

immutable VkClearColorValue
    _VkClearColorValue::Array_4_Cfloat
end

immutable VkClearDepthStencilValue
    depth::Cfloat
    stencil::UInt32
end
```

```julia
julia> sizeof(VkClearColorValue)
16

julia> sizeof(VkClearDepthStencilValue)
8
```

ihnorton:
Hmm, do you still want this? Does it work? We could merge it and put it under a feature flag or something if it would create issues for other users.

vchuravy:
I no longer have a use for this, but @SimonDanisch might have an opinion.

(I mostly just went theough my list of open PR's and closed the ones I no longer cared about)

PR #159, opened 2016-03-29
Handling literals, fixes #158

### LINKED ISSUE #158: Rewriting defines
Is there a way of rewriting defines? While wrapping vulkan I encounter the problem that a few constant defines are skipped. They are of the form

```C
#define VK_REMAINING_MIP_LEVELS           (~0U)
```
```julia
# Skipping MacroDefinition: VK_REMAINING_MIP_LEVELS ( ~ 0U )
```

vchuravy:
In order to fix #158 correctly I added `~` to the list of valid identifiers. But there is also type information present that is needed for `~` to work correctly. 

```C
#define VK_REMAINING_MIP_LEVELS           (~0U)
```

is now turned into:
```julia
const VK_REMAINING_MIP_LEVELS = ~(UInt32(0))
```

see: https://github.com/JuliaGPU/Vulkan.jl/pull/3/commits/d268b6b2022352a8409b9bc6e0f44e5849c39c01 for an example.

### DIFF
--- a/src/wrap_c.jl
+++ b/src/wrap_c.jl
@@ -529,7 +529,7 @@ end
 
 function handle_macro_exprn(tokens::TokenList, pos::Int)
     function trans(tok)
-        ops = ["+" "-" "*" ">>" "<<" "/" "\\" "%" "|" "||" "^" "&" "&&"]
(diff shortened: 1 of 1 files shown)


ihnorton:
:+1:

PR #161, opened 2016-04-05
make installation instructions more fool proof

SimonDanisch:
I got convinced that I need llvm 3.3 when using `julia 4.4` because of this line: `Note that Clang.jl must be built against the same LLVM version used in Julia`. Also installation alone isn't enough, the `PATH` variable has to be adapted as well!


ihnorton:
:+1: 

You and @vchuravy have commit access now. Please feel free to merge smaller stuff and do the usual merge-after-few-days thing for bigger changes.

PR #162, opened 2016-04-05
Better Enums [do not merge!]

SimonDanisch:
This emits `@cenum` for `EnumDecl`.
We need to resolve https://github.com/JuliaLang/julia/issues/15728 first, though.
We could make a package `CEnum` with this [code](https://gist.github.com/SimonDanisch/88f73018d6bf5219574ec1c679bc76ef) and then just make wrappers depend on it.

### DIFF
--- a/src/wrap_c.jl
+++ b/src/wrap_c.jl
@@ -362,25 +362,24 @@ function wrap(context::WrapContext, expr_buf::OrderedDict, cursor::EnumDecl; use
     if (usename == "" && (usename = name(cursor)) == "")
         usename = name_anon()
     end
-    enumname = usename
-
-    buf = Any[]
-    enum_exprs = ExprUnit(buf)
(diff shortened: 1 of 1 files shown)


visr:
Thanks, this is great. I already wanted to use this in GDAL.jl, but preferred to stick with normal Enums for now.

Two things I needed to get it working for me:

Set the Enum type:
https://github.com/visr/Clang.jl/commit/cda8331eb4803424882757ab3d004b6c98fe7987

And for the case you mentioned with two fields for one value I use this workaround:
https://github.com/visr/Clang.jl/commit/4e18317b76e742119d3fe7d03ec6559ab819d55e

ihnorton:
@visr I'm a little (well, a lot) behind on merging and updating things around here, but a surprising number of people still seem to use this package so I'd be happy to merge any of the general fixes you pointed to!

PR #164, opened 2016-05-14
Fix depwarn on 0.5 and unnecessary use of `eval`.

yuyichao:
Along with https://github.com/JuliaLang/DataStructures.jl/pull/202 the test should have no depwarn anyore =)


ihnorton:
👍  thanks. Guess we should probably tag a new version pretty soon.

PR #167, opened 2016-08-14
OSX: Let Homebrew install llvm automatically

### LINKED ISSUE #165: LoadError: failed process: Process(`which llvm-config`, ProcessExited(1)) [1]
Hello,

I'm trying to build Clang under Mac OS X 10.10.5

```
julia> Pkg.build("Clang")
INFO: Building Clang
================================[ ERROR: Clang ]================================

LoadError: failed process: Process(`which llvm-config`, ProcessExited(1)) [1]
while loading /Users/scls/.julia/v0.4/Clang/deps/build.jl, in expression starting on line 12

================================================================================

==============================================[ BUILD ERRORS ]==============================================

WARNING: Clang had build errors.

 - packages with build errors remain installed in /Users/scls/.julia/v0.4
 - build the package(s) and all dependencies with `Pkg.build("Clang")`
 - build a single package by running its `deps/build.jl` script

============================================================================================================
```

```bash
$ llvm-config
bash: llvm-config: command not found
```

```bash
$ llvm-gcc --version
Apple LLVM version 7.0.2 (clang-700.1.81)
Target: x86_64-apple-darwin14.5.0
Thread model: posix
```

I don't know how to fix this.

Any idea ?

Kind regards

jiahao:
This PR is an attempt to fix #165 by using Homebrew.jl to install `llvm`
automatically in user space if `which llvm-config` does not pick up an
existing installation of LLVM.

The main change is to now load Homebrew.jl on OSX every time
`Pkg.build("Clang")` runs, and if no system installation of LLVM could
be found, uses a user space `llvm` keg (installing it as necessary).

Summary of other changes:

- Introduces Homebrew as an OSX-specific dependency
- Add short documentation of `find_llvm()` in `deps/build.jl`.
  `find_llvm()` now consistently returns an error if none of the
  `which llvm-config` invocations return useful information.
  Currently, `find_llvm()` can simply return `false` on the last line,
  which poisons the rest of the build script.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,15 +1,37 @@
-#
-
-find_llvm() = begin
+#Returns the path as determined by llvm-config, or throws an error if LLVM
+#could not be found
+function find_llvm()
     if haskey(ENV, "TRAVIS")
         return readchomp(`which llvm-config-3.4`)
     end
     tmp = joinpath(JULIA_HOME, "llvm-config")
     isfile(tmp) && return tmp
-    tmp = readchomp(`which llvm-config`)
+    #Get path from llvm-config
+    try
+        tmp = readchomp(`which llvm-config`)
+    end
     isfile(tmp) && return tmp
+    error("No useable llvm-config found")
 end
-if !haskey(ENV, "LLVM_CONFIG") ENV["LLVM_CONFIG"] = find_llvm() end
+
+#On OSX, install llvm automatically through Homebrew.jl if no llvm installation
+#is found
+@static if is_apple()
+    using Homebrew
+
+    try
+        find_llvm()
+    catch #Not installed on system
+        info("Installing llvm through Homebrew.jl")
+        if !Homebrew.installed("llvm")
+            Homebrew.add("llvm")
+        end
+    end
+end
+
+if !haskey(ENV, "LLVM_CONFIG")
+    ENV["LLVM_CONFIG"] = find_llvm()
+end
 
 # on travis and some other systems, the llvm lib may not be in the
 # default path.
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
-julia 0.4
+julia 0.5-
 DataStructures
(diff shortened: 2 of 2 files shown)


deps/build.jl
```diff
@@ -1,15 +1,37 @@
-#
-
-find_llvm() = begin
+#Returns the path as determined by llvm-config, or throws an error if LLVM
+#could not be found
+function find_llvm()
     if haskey(ENV, "TRAVIS")
         return readchomp(`which llvm-config-3.4`)
     end
     tmp = joinpath(JULIA_HOME, "llvm-config")
```
vchuravy: I think since v"0.5" `llvm-config` no longer is in `JULIA_PATH`, but is installed to `joinpath(JULIA_PATH, "..", "tools")`.

See how `Cxx` locates `llvm-config` https://github.com/Keno/Cxx.jl/blob/77ae9463a5307780f5c4abf959ebb620ed17ce2e/deps/build.jl#L25

yuyichao: Cxx.jl only supports source build AFAIK. `llvm-config` doesn't seem to be included in current binary package.

ihnorton: Afk. Feel free to merge if/when it works.

jiahao: It seems that the julia-0.5 build provided by @staticfloat's PPA is simply too old (20160408) and does not support the `@static` macro syntax. Will re-test when binaries get updated.

staticfloat: I found one problem with the nightly build so far, you can follow along [here](https://github.com/JuliaLang/julia/pull/18047).

tkelman: Use `language: julia` instead of the PPA.

jiahao:
Crap, I pushed to the wrong branch. Sorry

tkelman:
Were you going to revert this?

jiahao:
Yes, I probably should. It works locally but let's get Travis to work first.

PR #180, opened 2017-05-22
Restore an orphan commit from the v0.0.4 tag

tkelman:
Doing some scraping of metadata, found that the v0.0.4 tag of Clang is the below orphaned commit, it's not in the history of any branches and wasn't pushed as a git tag. I don't actually want anyone to merge this, but just so this commit can be checked out without getting a "fatal: reference is not a tree" error, it would be good if someone with commit access could do
```
git remote add tkelman https://github.com/tkelman/Clang.jl
git fetch tkelman
git checkout c4d18994e7b4f6a5d8de6f88eea681639528a74a
git tag v0.0.4
git push origin v0.0.4
```

That is all, if someone does that, we can go ahead and close this. Thanks!


ihnorton:
Thanks!

PR #184, opened 2017-07-15
WIP: Drop 0.5 support

ihnorton:


yuyichao:
There shouldn't be a need to drop 0.5 support (not that I personally care much....)

ihnorton:
I'll try to make sure the output is backwards compatible. Not going to bother testing on 0.5 otherwise though, so might as well say so.

PR #188, opened 2017-07-20
Fix LLVM install using Homebrew.jl on macOS

rsrock:


rsrock:
I'm not sure what happened with the Travis changes.  Adding Macs seemed to work over here: 
https://travis-ci.org/rsrock/Clang.jl/builds/255793312?utm_source=github_status&utm_medium=notification

deps/build.jl
```diff
@@ -51,7 +51,8 @@ function find_llvm()
                      Homebrew.add("llvm")
                    end
                  end)
-            if (lcf = isfile(joinpath(prefix, "llvm-config")))
+            if (lcf = isfile(joinpath(Homebrew.prefix(),
+                                      "deps/opt/llvm/llvm-config")))
```
Gnimuc: It seems the path has been changed again, this works on my machine:
```julia
lcf = joinpath(Homebrew.prefix(), "Cellar/llvm/6.0.0/bin/llvm-config")
isfile(lcf) && return lcf
```

vchuravy:
@ihnorton What parts of clang do you need Elliot and I are working on getting LLVM and Clang building with BinaryProvider (see https://github.com/staticfloat/LLVMBuilder) and we aim to provide Clang.

ihnorton:
Just libclang.so/dylib (C API).

Are you envisioning support for Cxx.jl too? Some things I've done in the past (and may try to do again) needed a few hooks in to the C++ API for functionality not exposed through libclang -- but much less of the API than Cxx.jl needs.

Gnimuc:
closed in favor of #216

PR #194, opened 2018-04-16
Rewrite wrapper to call libclang directly

ihnorton:
Tests pass, but still need to check on a few things.
Will make installation much simpler: no more compiling shim against libclang, only need libclang as a shared library.


Gnimuc:
bump, any update plan for 1.0?

ihnorton:
Getting closer. Passes tests for the libclang binding portion. wrap_c doesn't have very much coverage, so probably some work left there. I'll probably need to run some generators to shake that out.

ihnorton:
@stemann thanks btw, I pulled those commits in to this branch.

ihnorton:
Alright, tested on Sundials wrapper, and looks ok: runs all the way through, output has some differences but may be due to version bump or OS difference. I'm hoping to have time to dig a bit more and firm up the tests, but this is at least in the realm of usable now.

PR #201, opened 2018-09-14
Another upgrading oversight -- replace

Gnimuc:
maybe we could improve tests for capturing this?


ihnorton:
Thanks!

> maybe we could improve tests for capturing this?

👍

PR #210, opened 2018-10-11
RFC: Refactor the package to make it self-generated and more

Gnimuc:
I'm investigating how to rebuild the package directly on the auto-generated libclang bindings. Almost done. Please give it a review at your convenience ;)

## Change Summary
* rebuild `Clang.jl` directly on top of the auto-generated libclang C bindings
  - the new submodule `LibClang` is one-to-one mapped to the libclang C APIs. it depends on two new files: `CEnum.jl` and `ctypes.jl`. `CEnum.jl` is for better C-compatible enum handling. `ctypes.jl` contains those C stdlib types that are missing corresponding Julia aliases. when generating wrappers, we should also ship these two files to users.
  - module `cindex`'s base functions are now rebuild on top of LibClang. almost all of the old APIs are renamed, please take a look at `index.jl`, `trans_unit.jl`, `cursor.jl`, `type.jl`, `token.jl` and `show.jl` for details.
  - ~~module `cindex`'s CL-prefixed types are removed, since methods are now directly dynamic-dispatched on `Val{CEnum}` value types(not sure whether this is a good idea or not).~~
  - unwind `wrap_c.jl` and add an AbstractContext type for extensibility
    + this is easier for users who would like to write their own `wrap!` for certain cursor decls
    and dispatched on their own context.
    + for compatibility's sake, `init`+`run` are rebuild on top of `DefaultContext`. also, old `cindex` and `wrap_c` modules can be loaded with `Clang.Deprecated.cindex; using Clang.Deprecated.wrap_c;`
* provide an out-of-box installation experience with BB2 on Linux, masOS and Windows, this may attract more users to use this package.
  - the newly released binaries of `LLVMBuilder` are still not compatible with Julia-1.0, so I built a compatible version with an older version of `BinaryBuilder` locally and uploaded on the Github releases of this repo.
* bugfixes and TODO list:
- [x] handle typedef anonymous structs/enums correctly using visr's method(close #208, fix #190)
- [x] integrate CEnums(cherrypick Simon's old PR close #162) 
- [x] refactor cindex directly on top of generated wrapper - [x] make wrap_c work again 
- [x] add tests 
- [x] use LLVMBuilder to provide an out-of-box installation experience:
  - fix #211, fix #197, fix #183, fix #65 
  - potentially fix #163 and #39 since with BB2 we install those headers in the same place on all platforms
  - close #172(obsolete), close #145(obsolete), close #142(looks like an installation problem), close #75(obsolete), close #49(obsolete)
[...]

PR #216, opened 2018-12-23
#210

Gnimuc:
try #210 again


Gnimuc:
rebasing in gitkraken takes no effect, but manually rebasing in CLI works, not sure why.

PR #223, opened 2019-02-24
Support for anonymous nested record

### LINKED ISSUE #222: Support for `typedef union`
I have a fairly large and complicated union and `Clang.jl` produces:

```
struct ucm_event
    ::struct ucm_event::(anonymous at /home/vchuravy/src/UCX.jl/deps/usr/include/ucm/api/ucm.h:80:5)
end
```

original C-code:

```
/**
 * @brief Memory event parameters and result.
 */
typedef union ucm_event {
    /*
     * UCM_EVENT_MMAP
     * mmap() is called.
     * callbacks: pre, post
     */
    struct {
        void           *result;
        void           *address;
        size_t         size;
        int            prot;
        int            flags;
        int            fd;
        off_t          offset;
    } mmap;

    /*
     * UCM_EVENT_MUNMAP
     * munmap() is called.
     */
    struct {
        int            result;
        void           *address;
        size_t         size;
    } munmap;

    /*
     * UCM_EVENT_MREMAP
     * mremap() is called.
     */
    struct {
        void           *result;
        void           *address;
        size_t         old_size;
        size_t         new_size;
        int            flags;
    } mremap;

    /*
     * UCM_EVENT_SHMAT
     * shmat() is called.
     */
    struct {
        void           *result;
        int            shmid;
        const void     *shmaddr;
        int            shmflg;
    } shmat;

    /*
     * UCM_EVENT_SHMDT
     * shmdt() is called.
     */
    struct {
        int            result;
        const void     *shmaddr;
    } shmdt;

    /*
     * UCM_EVENT_SBRK
     * sbrk() is called.
     */
    struct {
        void           *result;
        intptr_t       increment;
    } sbrk;

    /*
     * UCM_EVENT_MADVISE
     * madvise() is called.
     */
    struct {
        int            result;
        void           *addr;
        size_t         length;
        int            advice;
    } madvise;

    /*
     * UCM_EVENT_VM_MAPPED, UCM_EVENT_VM_UNMAPPED
     *
     * This is a "read-only" event which is called whenever memory is mapped
     * or unmapped from process address space, in addition to the other events.
     * It can return only UCM_EVENT_STATUS_NEXT.
     *
     * For UCM_EVENT_VM_MAPPED, callbacks are post
     * For UCM_EVENT_VM_UNMAPPED, callbacks are pre
     */
    struct {
        void           *address;
        size_t         size;
    } vm_mapped, vm_unmapped;

    /*
     * memory type allocation and deallocation event
     */
    struct {
        void           *address;
        size_t         size;
        ucm_mem_type_t mem_type;
    } mem_type;

} ucm_event_t;
```

Gnimuc:
Since Julia doesn't support defining nested structs:
```julia
struct Foo
    struct Bar
         b
    end
    x::Bar  # x can't see Bar
end
```
the workaround used here is to define a global hard-coded mangling "anonymous" struct and then treat it as a common struct when generating ASTs. 

- [x] nested union fix #222 
- [x] nested struct
- [x] add warning message for #224


vchuravy:
I found two cases were this branch did mostly the right thing: https://github.com/JuliaInterop/Clang.jl/issues/222#issuecomment-466810333

Gnimuc:
For #224, should we generate code for `__attribute__` enums with a warning message or not generate code at all?

Gnimuc:
merged and tagged v0.8.2.

PR #229, opened 2019-04-01
Improve GC sanity

Gnimuc:
Remove the misuse of `GC.@preserve`.

### DIFF
--- a/gen/libclang_api.jl
+++ b/gen/libclang_api.jl
@@ -954,7 +954,7 @@ function clang_getIBOutletCollectionType(arg1)
 end
 
 function clang_visitChildren(parent, visitor, client_data)
-    ccall((:clang_visitChildren, libclang), UInt32, (CXCursor, CXCursorVisitor, Any), parent, visitor, client_data)
+    ccall((:clang_visitChildren, libclang), UInt32, (CXCursor, CXCursorVisitor, CXClientData), parent, visitor, client_data)
 end
 
 function clang_getCursorUSR(arg1)
@@ -1450,3 +1450,4 @@ function clang_Type_visitFields(T, visitor, client_data)
 end
 # Julia wrapper for header: Platform.h
 # Automatically generated using Clang.jl
+
--- a/gen/libclang_api.jl
+++ b/gen/libclang_api.jl
@@ -954,7 +954,7 @@ function clang_getIBOutletCollectionType(arg1)
 end
 
 function clang_visitChildren(parent, visitor, client_data)
-    ccall((:clang_visitChildren, libclang), UInt32, (CXCursor, CXCursorVisitor, CXClientData), parent, visitor, client_data)
+    ccall((:clang_visitChildren, libclang), UInt32, (CXCursor, CXCursorVisitor, Any), parent, visitor, client_data)
 end
 
 function clang_getCursorUSR(arg1)
@@ -1450,4 +1450,3 @@ function clang_Type_visitFields(T, visitor, client_data)
 end
 # Julia wrapper for header: Platform.h
 # Automatically generated using Clang.jl
-
--- a/src/cltypes.jl
+++ b/src/cltypes.jl
@@ -20,9 +20,9 @@ for sym in enum_names(CXCursorKind)
 end
 
 CLCursor(c::CXCursor) = CXCursorMap[c.kind](c)
-Base.convert(::Type{CXCursor}, x::T) where {T<:CLCursor} = x.cursor
-Base.convert(::Type{CLCursor}, x::CXCursor) = CLCursor(x)
-
+Base.convert(::Type{CXCursor}, x::CLCursor) = x.cursor
+Base.convert(::Type{T}, x::CXCursor) where {T<:CLCursor} = CLCursor(x)
+Base.unsafe_convert(::Type{CXCursor}, x::CLCursor) = x.cursor
 
 # each CXTypeKind enum gets a specific Julia type wrapping
 # so that we can dispatch directly on node kinds.
@@ -41,9 +41,9 @@ for sym in enum_names(CXTypeKind)
 end
 
 CLType(c::CXType) = CLTypeMap[c.kind](c)
-Base.convert(::Type{CXType}, x::T) where {T<:CLType} = x.type
-Base.convert(::Type{CLType}, x::CXType) = CLType(x)
-
+Base.convert(::Type{CXType}, x::CLType) = x.type
+Base.convert(::Type{T}, x::CXType) where {T<:CLType} = CLType(x)
+Base.unsafe_convert(::Type{CXType}, x::CLType) = x.type
 
 # CLToken
 abstract type CLToken end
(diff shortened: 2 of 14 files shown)


src/cursor.jl
```diff
@@ -369,38 +335,30 @@ isvirtual(c::CLCursor) = isvirtual(c.cursor)
 #
 
 """
-    spelling(c::CXCursor) -> String
-    spelling(c::CLCursor) -> String
+    spelling(c::Union{CXCursor,CLCursor}) -> String
 Return a name for the entity referenced by this cursor.
 """
-function spelling(c::CXCursor)
-    GC.@preserve c begin
```
yuyichao: This preserve seems necessary if either the two input types use finalizer to manage the memory related to any of the pointers

It would not be necessary if `cxstr` memory is not bound to `c`

src/cursor.jl
```diff
@@ -546,10 +494,8 @@ Return a child cursor vector of the given cursor.
 function children(cursor::CXCursor)
     # TODO: possible to use sizehint! here?
     list = CXCursor[]
-    GC.@preserve cursor begin
```
yuyichao: If you are following the normal convention then you need to preserve list but not cursor

src/cursor.jl
yuyichao: And same with most of the ones that preserve c below

Gnimuc: According to https://clang.llvm.org/doxygen/CXString_8h_source.html：
> when the ownership of that string might differ from one call to the next

I guess this implies `cxstr` memory is not bound to `c`.

src/cursor.jl
Gnimuc: Just to make sure I understood this correctly. It's necessary to preserve `list` because in the case when the return value of `children` is discarded, the compiler might optimize it out?

yuyichao: No. `list` can be freed whenever `pointer_from_objref(list)` returns.

yuyichao: Also, this code will be much better written as.

```
function cu_children_visitor(cursor::CXCursor, parent::CXCursor, list)::Cuint
    push!(list, cursor)
    return CXChildVisit_Continue
end

function children(cursor::CXCursor)
    list = CXCursor[]
    cu_visitor_cb = @cfunction(cu_children_visitor, Cuint, (CXCursor, CXCursor, Ref{Vector{CXCursor}}))
    GC.@preserve list begin; clang_visitChildren(cursor, cu_visitor_cb, pointer_from_objref(list)); end
    return list
end
```

Or even replacing `clang_visitChildren(cursor, cu_visitor_cb, pointer_from_objref(list))` with `ccall((:clang_visitChildren, libclang), UInt32, (CXCursor, CXCursorVisitor, Any), cursor, cu_visitor_cb, list)`

src/cltypes.jl
```diff
@@ -20,9 +20,9 @@ for sym in enum_names(CXCursorKind)
 end
 
 CLCursor(c::CXCursor) = CXCursorMap[c.kind](c)
-Base.convert(::Type{CXCursor}, x::T) where {T<:CLCursor} = x.cursor
```
yuyichao: These conversion rules seems suspicious. It's almost always wrong to define `unsafe_convert` to be the same as `convert`. Is the `convert` being used anywhere or is it just for `ccall`s?

src/cursor.jl
Gnimuc: > No. `list` can be freed whenever pointer_from_objref(list) returns.

`list` is the return variable of `children`. How could it be GC-collected before returned?

src/cltypes.jl
Gnimuc: Just for `ccall`s, it's being called through `Base.cconvert`. Should I overload `cconvert` to `Base.cconvert(::Type{CXCursor}, x::CLCursor) = x` instead?

src/cursor.jl
yuyichao: > `list` is the return variable of `children`

That doesn't matter. FWIW, even though the compiler should not do this ATM, the compiler can proof in this case that there's no valid use of `pointer_from_objref(list)` before the return therefore there's no use of `list` before the return and so it doesn't need to allocate `list` before the return at all. The return value may also not be used after inlining.

src/cltypes.jl
yuyichao: Yes, you should never define `convert` only to be used for `cconvert` (That's likely a pre-0.4 legacy). I don't remember what's the default behavior for cconvert but yes, that looks like the correct `cconvert` implementation if it's not already the default behavior.

gen/libclang_api.jl
```diff
@@ -954,7 +954,7 @@ function clang_getIBOutletCollectionType(arg1)
 end
 
 function clang_visitChildren(parent, visitor, client_data)
-    ccall((:clang_visitChildren, libclang), UInt32, (CXCursor, CXCursorVisitor, CXClientData), parent, visitor, client_data)
```
yuyichao: This changes the API, which is why the version I wrote inlined the `ccall`. This is certainly fine if this is an internal API and you've updated all the callers.

You can also check if other ones has similar pattern.

FWIW, this seems like a very thin wrapper and putting the ccall inline doesn't seem too bad to me...... just personal taste biased on the readability for ppl that didn't write the codde...

Gnimuc:
@yuyichao Thanks for taking the time to review this! The only thing that I'm still not crystal clear about is: `GC.@preserve` has a clear begin-end scope, and the non-`Ptr{Cvoid}` return variable, say x, of `Base.cconvert` is invisible to GC, I'm wondering when will it be visible to GC again? If there is no mechanism to make it be visible to GC, I suppose it will not be GC-collected until the end of the current scope?

src/cursor.jl
yuyichao: > Or even replacing `clang_visitChildren(cursor, cu_visitor_cb, pointer_from_objref(list))` with `ccall((:clang_visitChildren, libclang), UInt32, (CXCursor, CXCursorVisitor, Any), cursor, cu_visitor_cb, list)`

And after this you don't need the `GC.@preserve list` anymore.

yuyichao:
GC visibility is completely independent of the scope and `GC.@preserve` has no effect any variable inside the code block. Preserving of `cconvert` of `ccall` argument is a completely different mechanism which has a scope till the return of the C function.

Also see the inline reply about the unneeded `GC.@preserve`.

Gnimuc:
> Preserving of cconvert of ccall argument is a completely different mechanism which has a scope till the return of the C function.

Just found the [doc](https://docs.julialang.org/en/latest/devdocs/llvm/#Supporting-[ccall](@ref)-1) about the mechanism used in `ccall`. Maybe it's a good idea to cross-ref it in Calling C and Fortran Code chapter.

Gnimuc:
I'm merging this. The codecov complains about a diff hit loss which is probably due to the coverage.jl failed to detect newly added `Base.cconvert`.

PR #231, opened 2019-08-08
Don't require rewriter to mutate buffer

visr:
This follows how the transformation is applied for the common file: https://github.com/JuliaInterop/Clang.jl/blob/c2c0ea3a6068fe2de40957d31a189691e7cfb4bf/src/compat.jl#L147

I was returning a new buffer in my rewriter and got confused it would only do anything for the common file.

### DIFF
--- a/src/compat.jl
+++ b/src/compat.jl
@@ -128,7 +128,7 @@ function Base.run(wc::WrapContext, generate_template=true)
         end
 
         # apply user-supplied transformation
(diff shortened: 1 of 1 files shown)


Gnimuc:
Thanks! Do you need a new tag for these two changes?

visr:
Great, thanks. I can continue fine on master for now, since my packages don't rely on Clang directly anyway.

Issue #249, opened 2020-01-29
Struct/union failing due to overlapping cursor enum in libclang Index.h


wsphillips:
https://clang.llvm.org/doxygen/Index_8h_source.html#l02582

Both `CXCursor_FirstAttr` and `CXCursor_UnexposedAttr` map to `400` which appears to cause issues when wrapping certain C structs and unions.

Example header declarations:
```c
 #ifndef CVITime_DECLARED
 #define CVITime_DECLARED
         typedef struct CVITime { uInt64 lsb; int64 msb; } CVITime;
 #endif
 #ifndef CVIAbsoluteTime_DECLARED
 #define CVIAbsoluteTime_DECLARED
         typedef union CVIAbsoluteTime { CVITime cviTime; uInt32 u32Data[4]; } CVIAbsoluteTime;
 #endif
```

Resulting cursor info when parsed by `Clang.jl`:
```julia
using Clang, Clang.LibClang

ctx = DefaultContext()
trans_unit = parse_header("NIDAQmx.h")
root_cursor = getcursor(trans_unit)
struct_cursor = search(root_cursor, "CVITime")[1]

for c in children(struct_cursor)
    println("Cursor: ", c, "\n  Kind: ", kind(c), "\n  Name: ", name(c), "\n  Type: ", type(c))
end

wrap!(ctx, struct_cursor)
```

Output:
```
Cursor: CLCursor (CLUnexposedAttr)
  Kind: CXCursor_FirstAttr
  Name:
  Type: CLType (CLInvalid)
Cursor: CLCursor (CLFieldDecl) lsb
  Kind: CXCursor_FieldDecl
  Name: lsb
  Type: CLType (CLTypedef)
Cursor: CLCursor (CLFieldDecl) msb
  Kind: CXCursor_FieldDecl
  Name: msb
  Type: CLType (CLTypedef)
┌ Warning: Skipping struct: "CLCursor (CLStructDecl) CVITime" due to unsupported field: CLCursor (CLUnexposedAttr)
└ @ Clang ~/.julia/packages/Clang/CiPzM/src/wrap_c.jl:156
```

I have fixed this locally by skipping fields where `field_kind == CXCursor_FirstAttr` in the struct/union dispatches for `wrap!()`. However, I don't know whether this is a more general bug, or if my fix generalizes appropriately.

[...]

PR #250, opened 2020-01-29
(Possible) workaround for structs/unions that have unexposed FirstAttr fields

wsphillips:
Attempt at fix for issue #249

### DIFF
--- a/src/wrap_c.jl
+++ b/src/wrap_c.jl
@@ -151,6 +151,8 @@ function wrap!(ctx::AbstractContext, cursor::CLStructDecl)
         field_kind = kind(field_cursor)
         if field_kind == CXCursor_StructDecl || field_kind == CXCursor_UnionDecl
             continue
+        elseif field_kind == CXCursor_FirstAttr
+            continue
         elseif field_kind != CXCursor_FieldDecl || field_kind == CXCursor_TypeRef
             buffer[struct_sym] = ExprUnit(Poisoned())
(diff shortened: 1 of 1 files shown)


Gnimuc:
This makes me feel again that libclang is not an appropriate tool for building a wrapper generator. It looks like the major audience of libclang is IDE-tools-writer, that's why it doesn't expose enough APIs we needed and hence some subroutines of the `wrap!()` function look too hairy. 

I think one day we should rewrite the package to be a true wrapper over Clang's C++ API instead of libclang's C API. Cxx.jl is a very promising tool for this job, but it lacks good maintenance for now. I would expect this to be done in the Julia-v1.4~v1.5 period. But before that, I guess it's good to go with libclang + hairy code.

Gnimuc:
Better to add some tests in this PR. :)

wsphillips:
> This makes me feel again that libclang is not an appropriate tool for building a wrapper generator. It looks like the major audience of libclang is IDE-tools-writer, that's why it doesn't expose enough APIs we needed and hence some subroutines of the `wrap!()` function look too hairy.

Yeah, I've just adopted existing `wrap!()` routines as a starting point/template in most cases. To be fair, it's partly to do with sometimes poorly written C APIs!

I will be sure to come back and add tests in a couple days once I have this current mess out of the way.

Gnimuc:
I'm going to merge this, tests can be added in another PR in the future. Thanks for contribution!

PR #265, opened 2020-05-25
Use CBinding to support more complex codes and translate only required symbols.

phuchant:
- Use CBinding to support more complex C codes: structures with bitfields, variadic functions..
- Only translate symbols selected by the user and datatypes they depend on (determined automatically)


ihnorton:
@phuchant thanks, from a quick look this looks very interesting, and it would be great to support additional constructs. I haven't been involved here for a while now, so I'll defer to @Gnimuc first if they have time to review. It does look like there are some API changes (e.g. [here](https://github.com/JuliaInterop/Clang.jl/pull/265/files#diff-92942c0e13e33a2b673529556f61be3aR49)) where functions are expected -- that might break generators for some existing users -- usually when making larger changes I have tried in the past to run against a few existing users like Sundials (or others in the README), and ensure that their scripts still work and the output is reasonable.

PR #302, opened 2021-06-22
Remove "blacklist" and "whitelist" in favor of neutral terms

logankilpatrick:
I also have a PR open here: https://github.com/JuliaLang/SuiteSparse.jl/pull/36

### DIFF
--- a/src/generator/passes.jl
+++ b/src/generator/passes.jl
@@ -684,10 +684,10 @@ In this pass, the mutability of those structs which are not necessary to be immu
 will be reset to `true` according to the following rules:
 
 if this type is not used as a field type in any other types
-    if this type is in the whitelist
+    if this type is in the allowlist
         then reset
 
-    if this type is in the blacklist
+    if this type is in the denylist
         then skip
 
     if this type is used as the argument type in some function protos
@@ -710,8 +710,8 @@ function (x::TweakMutability)(dag::ExprDAG, options::Dict)
     general_options = get(options, "general", Dict())
     log_options = get(general_options, "log", Dict())
     show_info = get(log_options, "TweakMutability_log", x.show_info)
-    blacklist = get(general_options, "auto_mutability_blacklist", [])
-    whitelist = get(general_options, "auto_mutability_whitelist", [])
+    denylist = get(general_options, "auto_mutability_denylist", [])
+    allowlist = get(general_options, "auto_mutability_allowlist", [])
     add_new = get(general_options, "auto_mutability_with_new", true)
 
     # collect referenced node ids
@@ -737,9 +737,9 @@ function (x::TweakMutability)(dag::ExprDAG, options::Dict)
         type_name = string(expr.args[2])
 
         apply_reset = false
-        if type_name ∈ whitelist
+        if type_name ∈ allowlist
             apply_reset = true
-        elseif type_name ∈ blacklist
+        elseif type_name ∈ denylist
             apply_reset = false
         else
             apply_reset = should_tweak(dag.nodes, i)
@@ -802,12 +802,12 @@ function (x::FunctionPrinter)(dag::ExprDAG, options::Dict)
     general_options = get(options, "general", Dict())
     log_options = get(general_options, "log", Dict())
     show_info = get(log_options, "FunctionPrinter_log", x.show_info)
-    blacklist = get(general_options, "printer_blacklist", [])
+    denylist = get(general_options, "printer_denylist", [])
 
     show_info && @info "[FunctionPrinter]: print to $(x.file)"
     open(x.file, "w") do io
         for node in dag.nodes
-            string(node.id) ∈ blacklist && continue
+            string(node.id) ∈ denylist && continue
             node.type isa AbstractFunctionNodeType || continue
             pretty_print(io, node, general_options)
         end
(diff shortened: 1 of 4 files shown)


Gnimuc:
Sorry, I'm not a native English speaker, and I don't know what's the point here. Is it ok to support both of those options? so we don't need to update those scripts in the downstream packages.

johnnychen94:
I'm against this not because it's not "right", I'm against this because I don't think it worths the effort; people here are already busy enough to deal with multiple packages. If it's dealing with new code, 100% great. If it's renaming the old codes in a breaking way, that's not a good option.

If we want to recommend these new terminology, I think what's needed to be done first is to update the community standards first.

logankilpatrick:
@gnimuc the point is to move away from terms intertwined with race and racism (I.e. whiteness being "good" thus is allowed and blackness being "bad") happy to PR the CoC as well.

KristofferC:
What do white and black have to do with race? This is pretty much a purely American thing in that you guys have separated up your population into these two distinct classes. If anything the problematic part. Not that the word "black" is used for the absence of something where the obvious origin, lack of light, or inability to reflect light, is used everywhere all over the place (black hole, dark matter, blackout, etc etc).

logankilpatrick:
Not sure this PR is the right place for a discussion like this @KristofferC happy to chat elsewhere

KristofferC:
Sure, it is kinda problematic to open PRs where the content of them cannot be discussed therein though.

logankilpatrick:
For some additional context: https://www.zdnet.com/article/uk-ncsc-to-stop-using-whitelist-and-blacklist-due-to-racial-stereotyping/, https://9to5google.com/2020/06/12/google-android-chrome-blacklist-blocklist-more-inclusive/, https://thenextweb.com/news/github-plans-to-replace-racially-insensitive-terms-like-master-and-whitelist, etc. 

Not sure it's fair to say that the terms white or black are a US thing, it seems pretty generally accepted from leaders in the tech space that when at all possible, we should move away from terms like this. I have no appetite to argue about this though.

oxinabox:
> I'm against this not because it's not "right", I'm against this because I don't think it worths the effort; people here are already busy enough to deal with multiple packages. If it's dealing with new code, 100% great. If it's renaming the old codes in a breaking way, that's not a good option.


As far as I can tell this PR is a purely internal change. It is completely invisible to the user.
The only effort to be spent is to review it.
And it is a pretty easy review.
The time taken to read this comment is about as long as it will take to review it.

oxinabox:
When renaming to avoid these terms it is also worth aiming to end up with something more clear.
As such this change is best done by someone experienced with the code-base (not me).
It seems like many of the places that say `blacklist` would be clearer being called 
`files_to_skip`, `always_treat_as_mutable`, `never_treat_as_mutable`, `treat_as_function`,
`forbidden_idks`.

To make this nonbreaking, one should introduce a way of using the removed keys still.
Something like the following works, though alternatives exist.
```julia
function get_with_deprecation(collection, new_name, old_name, default; combine=append!)
    haskey(collection, old_name) && Base.depwarn("the key $old_name has been deprecated in favour of $new_name", :general_ops)
    return combine(get(collection, new_name, default), get(collection, old_name, default))
end
```

comath:
Forgive me for I'm from outside the julia community, but I wanted to give my perspective. I work in security and the we make heavy use of allow and deny lists. So I've seen the switch happen elsewhere. Basically, @oxinabox brings up a very good point, and I think that approach has lead to better code than reusing using the same overloaded terms everywhere. When done well it'll work like curb cuts, meant to help some people but end up helping everyone.

prrao87:
@KristofferC I'd like to point out that along with everything @logankilpatrick has said, the term 'blacklist' originated in colonial times to deny certain people (mostly the colonized, who were people of colour), access to various things. It's definitely not reducible to being just an 'American thing'. It's 2021 now, and we can all agree that we want to live in a world where those folks whose ancestors might have been through so much, also feel just as included in the communitybas those to whom this issue doesn't matter as much.

A lot of other languages and platforms have made the switch, and it would be great for the Julia exosystem to become more supportive of the nomenclature in general. Granted, it causes breaking changes for a lot of downstream packages and is a LOT of work for maintainers, but that's a totally separate issue from the one where we fail to recognize that we can all do better.

johnnychen94:
> https://www.zdnet.com/article/linux-team-approves-new-terminology-bans-terms-like-blacklist-and-slave/
>
> > The Linux team did not recommend any specific terms but asked developers to choose as appropriate.
> >
> > The new terms are to be used for new source code written for the Linux kernel and its associated documentation.
> >
> > The older terms, considered inadequate now, will only be allowed for maintaining older code and documentation, or "when updating code for an existing (as of 2020) hardware or protocol specification that mandates those terms."

It's very clear to me that blacklist/whitelist belongs to the "older term" category and it's somehow breaking to update it to the new terminology. And it's very clear to me that none of us intends to associate it with race.

kescobo:
> because I don't think it worths the effort; people here are already busy enough to deal with multiple packages.

@johnnychen94 Good that you raised the issue of cost and value - no one would argue that a change like this has no cost, the question is really "is the cost worth it?" And part of that calculation is what is the cost of keeping things the same. I gather @KristofferC thinks that cost is minimal, for example, so it makes sense that he wouldn't thing the change is worth it.

It won't be me paying the maintenance cost, and I'm not someone that's adversely affected by the old language, so I don't have much of a vote here, but it seems to me that, regardless of the origins of the words, and regardless of anyone's current intent when using them, if a bunch of communities and companies **are** paying those maintenance costs for the sake of being more welcoming, and others aren't because they don't think it's worth the effort, some future black developer might draw a reasonable inference that the latter groups aren't as interested in their contributions.

amirali7e:
Is it just me or does this sound too ridiculous? What does blacklist and whitelist has anything to do with racism? What will it solve of racism? 
I know your intentions are decent and comes from a morality standpoint but please do not introduce unnecessary unrelated topics to the Julia community. We are here to do one thing perfectly as the community has been doing, we are not here to do political correctness and appease the woke trends.
Sure if there is a big backer that demands certain changes people must discuss it.

src/generator/passes.jl
```diff
@@ -802,12 +802,12 @@ function (x::FunctionPrinter)(dag::ExprDAG, options::Dict)
     general_options = get(options, "general", Dict())
     log_options = get(general_options, "log", Dict())
     show_info = get(log_options, "FunctionPrinter_log", x.show_info)
-    blacklist = get(general_options, "printer_blacklist", [])
+    denylist = get(general_options, "printer_denylist", [])
```
jedbrown, line 805:
```suggestion
    denylist = get(general_options, "printer_denylist", get(general_options, "printer_blacklist", []))
```
This would be a simple, local strategy to continue supporting the old names while allowing projects to transition to the new names. A specific deprecation could be added in the future, but at least this avoids throwing the unwelcoming flag of forcing other projects to write new dependencies on this identifier.

PR #303, opened 2021-06-26
Add pass for generating API methods with a function pointer parameter

serenity4:
Following from JuliaGPU/VulkanCore.jl#44, here is a pass for adding extra methods to API functions that `ccall` into libraries, exposing a function pointer argument for short-circuiting the library symbol lookup.

I tested it on VulkanCore.jl, both with `use_ccall_macro` turned on and off, and it seems to work as expected (update PR to come after this is merged). The test infrastructure did not seem to provide a handy way to test the behavior of this specific pass.

I made the inclusion of the pass to be a general option named "add_fptr_methods", feel free to suggest another name if that is not clear.

### DIFF
--- a/src/generator/passes.jl
+++ b/src/generator/passes.jl
@@ -771,33 +771,24 @@ function (::AddFPtrMethods)(dag::ExprDAG, options::Dict)
     use_ccall_macro = get(codegen_options, "use_ccall_macro", false)
     for node in dag.nodes
         node.type isa FunctionProto || continue
-        for _ex in copy(node.exprs)
-            ex = deepcopy(_ex)
-            Base.is_expr(ex, :function) || continue
-            if Base.is_expr(ex.args[1], :where)
-                call = ex.args[1].args[1]
-            else
-                call = ex.args[1]
-            end
-            push!(call.args, :fptr)
-            body_idx = findfirst(x -> Base.is_expr(x, :block), ex.args)
-            body = ex.args[body_idx]
-            idx = if use_ccall_macro
-                findfirst(x -> Base.is_expr(x, :macrocall) && x.args[1] == Symbol("@ccall"), body.args)
-            else
-                findfirst(x -> Base.is_expr(x, :call) && x.args[1] == :ccall, body.args)
-            end
-            !isnothing(idx) || continue
-            stmt = body.args[idx]
-            if use_ccall_macro
-                assert_idx = findfirst(x -> Base.is_expr(x, :(::)), stmt.args)
-                call_idx = findfirst(x -> Base.is_expr(x, :call), stmt.args[assert_idx].args)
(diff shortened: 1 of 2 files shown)


src/generator/passes.jl
```diff
@@ -756,6 +756,51 @@ function (x::TweakMutability)(dag::ExprDAG, options::Dict)
...
+struct AddFPtrMethods <: AbstractPass end
+
+function (::AddFPtrMethods)(dag::ExprDAG, options::Dict)
+    codegen_options = get(options, "codegen", Dict())
+    use_ccall_macro = get(codegen_options, "use_ccall_macro", false)
+    for node in dag.nodes
+        node.type isa FunctionProto || continue
+        for _ex in copy(node.exprs)
+            ex = deepcopy(_ex)
+            Base.is_expr(ex, :function) || continue
```
serenity4, line 776: Not sure if there would actually be expressions for nodes of type `FunctionProto` that are not functions.

src/generator/passes.jl
```diff
@@ -756,6 +756,51 @@ function (x::TweakMutability)(dag::ExprDAG, options::Dict)
...
+            end
+            push!(call.args, :fptr)
+            body_idx = findfirst(x -> Base.is_expr(x, :block), ex.args)
+            body = ex.args[body_idx]
+            idx = if use_ccall_macro
+                findfirst(x -> Base.is_expr(x, :macrocall) && x.args[1] == Symbol("@ccall"), body.args)
+            else
+                findfirst(x -> Base.is_expr(x, :call) && x.args[1] == :ccall, body.args)
+            end
+            !isnothing(idx) || continue
```
serenity4, line 790: I don't know if there are methods for `FunctionProto` nodes that don't `ccall` into libraries. If that's never the case then we can remove this line, which would otherwise be a silent failure.

serenity4 (commented):
Left a few comments regarding potential guarantees that we might have for the expressions contained in `FunctionProto` nodes.

serenity4:
To see the behavior, feel free to re-run the generator in VulkanCore.jl with this branch and setting a general option "add_fptr_methods" to true.

Gnimuc (changes requested):
Thanks for adding this feature! I re-run the generator script for VulkanCore.jl and the results look good: https://github.com/Gnimuc/GeneratorScripts/commit/627b412f734bdd5c1a37a48aa590fc94e3c98b86

Gnimuc:
> I made the inclusion of the pass to be a general option named "add_fptr_methods", feel free to suggest another name if that is not clear.

Looks like VulkanCore.jl is the only user of this option, so let's keep it secret for now and bikeshedding the name in another issue.

serenity4:
Thanks for the quick review. I made the suggested changes, it's a bit simpler now.

Gnimuc (approved):

Gnimuc:
Thanks for the contribution!

Issue #304, opened 2021-06-27
Constructors for union structs


serenity4:
The current master generates union structs like so:

```julia
struct VkClearColorValue
    data::NTuple{16, UInt8} # can hold homogeneous 4-tuples with type uint32, int32 or float32
end
```

where the `data` field is reinterpreted when accessed to match C unions.

As mentioned in [this comment](https://github.com/JuliaGPU/VulkanCore.jl/issues/40#issuecomment-792282988), because union structs hold raw bytes, it's not so convenient to instantiate them so we could define a constructor for union structs that help populate their bytes. We could generate something very similar to the snippet in the comment:

```julia
const __U_VkClearColorValue  = Union{NTuple{4,Float32},NTuple{4,Int32},NTuple{4,UInt32}}
VkClearColorValue(data::__U_VkClearColorValue) = VkClearColorValue(reinterpret(NTuple{16, UInt8}, [data])[])
```

So I am thinking of extending [this method](https://github.com/JuliaInterop/Clang.jl/blob/d691ad7af2e39b13e5989f8ccf6035c0f7e4e72c/src/generator/codegen.jl#L427) to emit the constructor method along with the `getproperty`/`setproperty!` methods. Would you accept a PR for adding this feature?

Gnimuc:
I think this is a special case where all 3 fields have the same size, and the size is equal to the size of the union. 

```
struct VkClearColorValue
    data::NTuple{16, UInt8}
end

function Base.getproperty(x::Ptr{VkClearColorValue}, f::Symbol)
    f === :float32 && return Ptr{NTuple{4, Cfloat}}(x + 0)
    f === :int32 && return Ptr{NTuple{4, Int32}}(x + 0)
    f === :uint32 && return Ptr{NTuple{4, UInt32}}(x + 0)
    return getfield(x, f)
end
```

Gnimuc:
By using the current generated field access methods, the simplest way to init a `VkClearColorValue` on the Julia side would be:

```
julia> x = Base.unsafe_convert(Ptr{VkClearColorValue}, Ref{VkClearColorValue}())
Ptr{VkClearColorValue} @0x000000013c2db190

julia> x.float32 = NTuple{4,Cfloat}((1,2,3,4))
(1.0f0, 2.0f0, 3.0f0, 4.0f0)

julia> unsafe_load(x)
[...]

PR #305, opened 2021-06-27
Generate constructors for union and struct records

serenity4:
Implements constructors for union structs, as discussed in #304, and also for struct records (in a very similar way).

I am unsure about struct records since fields may overlap, so any suggestions are welcome.

Because we use `setproperty!` on the pointer, which does most of the layout logic, I think this takes care of all possibilities including unions with offsets/different sizes, records with and without byte-aligned fields, records with unions, unions with records.

See below for a few examples of what the generated constructor looks like.

Union struct:
```julia
# definition
struct VkClearValue
    data::NTuple{16, UInt8}
end

# to see the layout
function Base.getproperty(x::Ptr{VkClearValue}, f::Symbol)
    f === :color && return Ptr{VkClearColorValue}(x + 0)
    f === :depthStencil && return Ptr{VkClearDepthStencilValue}(x + 0)
    return getfield(x, f)
end

# generated code from this PR

const __U_VkClearValue = Union{VkClearColorValue, VkClearDepthStencilValue}

function VkClearValue(val::__U_VkClearValue)
    ref = Ref{VkClearValue}()
    ptr = Base.unsafe_convert(Ptr{VkClearValue}, ref)
    if val isa VkClearColorValue
        ptr.color = val
    elseif val isa VkClearDepthStencilValue
        ptr.depthStencil = val
    end
    ref[]
end
```

Struct record:

```julia
# definition
struct VkAccelerationStructureInstanceKHR
    data::NTuple{64, UInt8}
end

# to see the layout
function Base.getproperty(x::Ptr{VkAccelerationStructureInstanceKHR}, f::Symbol)
    f === :transform && return Ptr{VkTransformMatrixKHR}(x + 0)
    f === :instanceCustomIndex && return Ptr{UInt32}(x + 48)
    f === :mask && return Ptr{UInt32}(x + 51)
    f === :instanceShaderBindingTableRecordOffset && return Ptr{UInt32}(x + 52)
    f === :flags && return Ptr{VkGeometryInstanceFlagsKHR}(x + 55)
    f === :accelerationStructureReference && return Ptr{UInt64}(x + 56)
    return getfield(x, f)
end

# generated code

function VkAccelerationStructureInstanceKHR(transform::VkTransformMatrixKHR, instanceCustomIndex::UInt32, mask::UInt32, instanceShaderBindingTableRecordOffset::UInt32, flags::VkGeometryInstanceFlagsKHR, accelerationStructureReference::UInt64)
    ref = Ref{VkAccelerationStructureInstanceKHR}()
    ptr = Base.unsafe_convert(Ptr{VkAccelerationStructureInstanceKHR}, ref)
    ptr.transform = transform
    ptr.instanceCustomIndex = instanceCustomIndex
    ptr.mask = mask
    ptr.instanceShaderBindingTableRecordOffset = instanceShaderBindingTableRecordOffset
    ptr.flags = flags
    ptr.accelerationStructureReference = accelerationStructureReference
    ref[]
end
```

### DIFF
--- a/src/generator/codegen.jl
+++ b/src/generator/codegen.jl
@@ -320,6 +320,73 @@ function emit_setproperty!(dag, node, options)
     return dag
 end
 
+function get_names_types(root_cursor, cursor, options)
+    field_cursors = fields(getCursorType(cursor))
+    field_cursors = isempty(field_cursors) ? children(cursor) : field_cursors
+    tys = []
+    fsyms = []
+    for field_cursor in field_cursors
+        n = name(field_cursor)
+        if isempty(n)
+            _emit_getproperty_ptr!(root_cursor, field_cursor, options)
+            continue
+        end
+        fsym = make_symbol_safe(n)
+        fty = getCursorType(field_cursor)
+        ty = translate(tojulia(fty), options)
+        push!(tys, ty)
+        push!(fsyms, fsym)
+    end
+    fsyms, tys
+end
+
+function emit_constructor!(dag, node::ExprNode{<:AbstractUnionNodeType}, options)
+    sym = make_symbol_safe(node.id)
+    fsyms, tys = get_names_types(node.cursor, node.cursor, options)
+    union_sym = Symbol(:__U_, sym)
+    push!(node.exprs, :(const $union_sym = Union{$(tys...)}))
+    body = Expr(:block,
+        :(ref = Ref{$sym}()),
+        :(ptr = Base.unsafe_convert(Ptr{$sym}, ref)),
+    )
+
+    branch_args = map(zip(fsyms, tys)) do (fsym, ty)
+        cond = :(val isa $ty)
+        assign = :(ptr.$fsym = val)
+        (cond, assign)
+    end
+
+    first_args, rest = Iterators.peel(branch_args)
+    ex = Expr(:if, first_args...)
+    push!(body.args, ex)
+    foreach(rest) do (cond, assign)
+        _ex = Expr(:elseif, cond, assign)
+        push!(ex.args, _ex)
+        ex = _ex
+    end
+
+    push!(body.args, :(ref[]))
+
+    push!(node.exprs, Expr(:function, Expr(:call, sym, :(val::$union_sym)), body))
+end
+
+function emit_constructor!(dag, node::ExprNode{<:StructLayout}, options)
+    sym = make_symbol_safe(node.id)
+    fsyms, tys = get_names_types(node.cursor, node.cursor, options)
+    body = quote
+        ref = Ref{$sym}()
+        ptr = Base.unsafe_convert(Ptr{$sym}, ref)
+        $((:(ptr.$fsym = $fsym) for fsym in fsyms)...)
+        ref[]
+    end
+
+    rm_line_num_node!(body)
+
+    push!(body.args, )
+    func = Expr(:function, Expr(:call, sym, (:($fsym::$ty) for (fsym, ty) in zip(fsyms, tys))...), body)
+    push!(node.exprs, func)
+end
+
 function emit!(dag::ExprDAG, node::ExprNode{<:AbstractStructNodeType}, options::Dict; args...)
     struct_sym = make_symbol_safe(node.id)
     block = Expr(:block)
@@ -434,6 +501,8 @@ function emit!(dag::ExprDAG, node::ExprNode{<:RecordLayouts}, options::Dict; arg
     emit_getproperty_ptr!(dag, node, options)
     emit_getproperty!(dag, node, options)
     emit_setproperty!(dag, node, options)
+
+    emit_constructor!(dag, node, options)
 
     return dag
 end
--- a/src/generator/codegen.jl
+++ b/src/generator/codegen.jl
@@ -382,7 +382,6 @@ function emit_constructor!(dag, node::ExprNode{<:StructLayout}, options)
 
     rm_line_num_node!(body)
 
-    push!(body.args, )
     func = Expr(:function, Expr(:call, sym, (:($fsym::$ty) for (fsym, ty) in zip(fsyms, tys))...), body)
     push!(node.exprs, func)
 end
--- a/src/generator/codegen.jl
+++ b/src/generator/codegen.jl
@@ -350,24 +350,37 @@ function emit_constructor!(dag, node::ExprNode{<:AbstractUnionNodeType}, options
         :(ptr = Base.unsafe_convert(Ptr{$sym}, ref)),
     )
 
-    branch_args = map(zip(fsyms, tys)) do (fsym, ty)
-        cond = :(val isa $ty)
-        assign = :(ptr.$fsym = val)
-        (cond, assign)
+    if get(options, "union_single_constructor", false)
+        branch_args = map(zip(fsyms, tys)) do (fsym, ty)
+            cond = :(val isa $ty)
+            assign = :(ptr.$fsym = val)
+            (cond, assign)
+        end
+
+        first_args, rest = Iterators.peel(branch_args)
+        ex = Expr(:if, first_args...)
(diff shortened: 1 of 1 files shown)


serenity4:
I think the idea is actually right for struct records, but the `setproperty!`/`getproperty` methods can be modified to not overflow into other fields for both getting and setting a field. The following works as intended:

```julia
struct VkAccelerationStructureInstanceKHR
    data::NTuple{64, UInt8}
end

# no changes
function Base.getproperty(x::Ptr{VkAccelerationStructureInstanceKHR}, f::Symbol)
    f === :transform && return Ptr{VkTransformMatrixKHR}(x + 0)
    f === :instanceCustomIndex && return Ptr{UInt32}(x + 48)
    f === :mask && return Ptr{UInt32}(x + 51)
    f === :instanceShaderBindingTableRecordOffset && return Ptr{UInt32}(x + 52)
    f === :flags && return Ptr{VkGeometryInstanceFlagsKHR}(x + 55)
    f === :accelerationStructureReference && return Ptr{UInt64}(x + 56)
    return getfield(x, f)
end

function unsafe_load_overlapping(ptr, nbytes)
    T = eltype(ptr)
    bytes = Base.unsafe_convert(Ptr{UInt8}, ptr)
    arr = zeros(UInt8, sizeof(T))
    arr[1:nbytes] .= unsafe_wrap(Array, bytes, nbytes)
    Base.reinterpret(T, arr)[]
end

function Base.getproperty(x::VkAccelerationStructureInstanceKHR, f::Symbol)
    r = Ref{VkAccelerationStructureInstanceKHR}(x)
    ptr = Base.unsafe_convert(Ptr{VkAccelerationStructureInstanceKHR}, r)
    fptr = getproperty(ptr, f)
    begin
        # change here; I didn't do the other branch but we could employ a similar logic
        if fptr isa Ptr
            GC.@preserve r begin
                # use unsafe_load_overlapping only for fields that overlap with others
                # (all the following fields are UInt32, so 4 bytes in size)
                f === :instanceCustomIndex && return unsafe_load_overlapping(fptr, 3)
                f === :mask && return unsafe_load_overlapping(fptr, 1)
                f === :instanceShaderBindingTableRecordOffset && return unsafe_load_overlapping(fptr, 3)
                f === :flags && return unsafe_load_overlapping(fptr, 1)
                return unsafe_load(fptr)
            end
        else
            (baseptr, offset, width) = fptr
            ty = eltype(baseptr)
            i8 = GC.@preserve(r, unsafe_load(baseptr))
            bitstr = bitstring(i8)
            sig = bitstr[(end - offset) - (width - 1):end - offset]
            zexted = lpad(sig, 8 * sizeof(ty), '0')
            return parse(ty, zexted; base = 2)
        end
    end
end

# this one was reworked a bit. Instead of using `unsafe_store!` (which stores all the bytes at once)
# we use `unsafe_copyto!` to store just the number of bytes we need (truncating in case of overflow)
# behavior is modified only for fields that overlap with others
function Base.setproperty!(x::Ptr{VkAccelerationStructureInstanceKHR}, f::Symbol, v)
    vref = Ref(v)
    GC.@preserve vref begin
        vptr = Base.unsafe_convert(Ptr{eltype(vref)}, vref)
        vbytes = Base.unsafe_convert(Ptr{UInt8}, vptr)
        xbytes = Base.unsafe_convert(Ptr{UInt8}, getproperty(x, f))
        if f === :transform
            unsafe_copyto!(xbytes, vbytes, 48)
        elseif f === :instanceCustomIndex
            unsafe_copyto!(xbytes, vbytes, 3)
        elseif f === :mask
            unsafe_copyto!(xbytes, vbytes, 1)
        elseif f === :instanceShaderBindingTableRecordOffset
            unsafe_copyto!(xbytes, vbytes, 3)
        elseif f === :flags
            unsafe_copyto!(xbytes, vbytes, 1)
        elseif f === :accelerationStructureReference
            unsafe_copyto!(xbytes, vbytes, 8)
        end
    end
end

# no changes from the above comment
function VkAccelerationStructureInstanceKHR(transform::VkTransformMatrixKHR, instanceCustomIndex::UInt32, mask::UInt32, instanceShaderBindingTableRecordOffset::UInt32, flags::VkGeometryInstanceFlagsKHR, accelerationStructureReference::UInt64)
    ref = Ref{VkAccelerationStructureInstanceKHR}()
    ptr = Base.unsafe_convert(Ptr{VkAccelerationStructureInstanceKHR}, ref)
    ptr.transform = transform
    ptr.instanceCustomIndex = instanceCustomIndex
    ptr.mask = mask
    ptr.instanceShaderBindingTableRecordOffset = instanceShaderBindingTableRecordOffset
    ptr.flags = flags
    ptr.accelerationStructureReference = accelerationStructureReference
    ref[]
end
```

serenity4:
Maybe we should tackle those changes in `getproperty/setproperty!` in a separate PR, since it may be a bit involved implementation-wise. For this PR you can skip the previous comment, since I haven't implemented the changes.

src/generator/codegen.jl
```diff
@@ -320,6 +320,73 @@ function emit_setproperty!(dag, node, options)
...
+    body = quote
+        ref = Ref{$sym}()
+        ptr = Base.unsafe_convert(Ptr{$sym}, ref)
+        $((:(ptr.$fsym = $fsym) for fsym in fsyms)...)
+        ref[]
+    end
+
+    rm_line_num_node!(body)
+
+    push!(body.args, )
```
Gnimuc, line 385: typo

Gnimuc:
`ref = Ref{VkClearValue}()` implicitly assumes that `VkClearValue` is an immutable isbits type. The `TweakMutability` pass is the only pass that can tweak the mutability, and it doesn't work for any memory-pool-like structures, so this assumption is always valid. 

However, depending on the API of the C library, it may be unnecessary to generate constructors for all of these memory-pool-like structures, so I believe we'd better make this feature optional by adding an option like:

https://github.com/JuliaInterop/Clang.jl/blob/d691ad7af2e39b13e5989f8ccf6035c0f7e4e72c/gen/generator.toml#L135-L136

Gnimuc:
> However I recognize that it would be more straightforward (at least from a readability perspective) to use multiple dispatch, so that the branching part gets removed. I don't have a preference, so I can go with the multiple dispatch style if you prefer.

If you don't mind, we can support both of these two styles by adding another option. 😜 

> We could make the option true by default (when not specified), and potentially have it specified as a list of excluded types. It can be the opposite if you think that this feature will be disabled more often than it will be enabled.

I'd say we keep it secret in this PR just like what we did in #303. I also think we should make this feature enabled by default in the final 0.14 release, but let's make the decision later, in another issue, and hear the voice from other users.

> We could call it something like `blacklist_record_layout_constructors`.

Better not use the word "blacklist", see #302. :trollface:

serenity4:
I ended up making an option `add_record_constructors` that can be set to `true`, `false` (default) or a list with specific symbols.

I also made an option for whether or not to generate a single constructor for unions (surprisingly named `union_single_constructor`) that is turned off by default. The generated code is identical as before if it is on, and if it is off we have multiple constructors defined as follows:

```julia
const __U_VkClearValue = Union{VkClearColorValue, VkClearDepthStencilValue}

function VkClearValue(color::VkClearColorValue)
    ref = Ref{VkClearValue}()
    ptr = Base.unsafe_convert(Ptr{VkClearValue}, ref)
    ptr.color = color
    ref[]
end

function VkClearValue(depthStencil::VkClearDepthStencilValue)
    ref = Ref{VkClearValue}()
    ptr = Base.unsafe_convert(Ptr{VkClearValue}, ref)
    ptr.depthStencil = depthStencil
    ref[]
end
```

Note that the constant is not used, but it may be useful for uses that want to dispatch on it, and I think it's better to have the same symbols regardless of the options we chose.

Gnimuc:
Thanks for the PR!

PR #312, opened 2021-07-14
Implement bitfield

melonedo:
The methods `getproperty` and `setproperty!` have become too long, should I turn them into prologue?
Also, are there methods to test the generated code?


Gnimuc:
hi @jpsamaroo, could you give this PR a try on your package https://github.com/jpsamaroo/BPFnative.jl?

Gnimuc:
The current implementation in the master branch is not good, so this is never a regression. Although this is not tested in any real-world projects, I think it's good to merge.

PR #318, opened 2021-07-24
Extract and dump C comments as documentation

melonedo:
- [x] Extract comments.
- [x] Remove comment markers e.g. `/**/` and `//`, doxygen comments are also supported.
- [x] Print comments for top-level nodes e.g. function, structs, enums.
- [x] Print comments for struct and enum members.
- [x] Support libclang's doxygen API.
- [x] Basic support of doxygen semantic, e.g. `\brief`.


src/cursor.jl
```diff
@@ -18,6 +18,14 @@ Base.hash(x::CLCursor) = hash(x.cursor)
 Base.:(==)(c1::CLCursor, c2::CXCursor) = c1.cursor == c2
 Base.:(==)(c1::CXCursor, c2::CLCursor) = c1 == c2.cursor
 
+function Base.convert(::Type{String}, cxstr::CXString)
+    ptr = clang_getCString(cxstr)
+    ptr == C_NULL && return ""
+    s = unsafe_string(ptr)
+    clang_disposeString(cxstr)
+    return s
+end
```
Gnimuc, line 27: This implementation is a bit unsafe. What will happen when someone calls this function twice? The behavior may depend on the implementation of `clang_disposeString`.

This is the implementation of `clang_disposeString`:
```
void clang_disposeString(CXString string) {
  switch ((CXStringFlag) string.private_flags) {
    case CXS_Unmanaged:
      break;
    case CXS_Malloc:
      if (string.data)
        free(const_cast<void *>(string.data));
      break;
    case CXS_StringBuf:
      static_cast<cxstring::CXStringBuf *>(
          const_cast<void *>(string.data))->dispose();
      break;
  }
}
```
It just frees the underlying resources without marking the flag field of the input `CXString` variable. 

So, ...

src/cursor.jl
```diff
@@ -394,8 +402,8 @@ isVariadic(c::Union{CXCursor,CLCursor}) = clang_Cursor_isVariadic(c) != 0
 
 # clang_Cursor_isExternalSymbol
 # clang_Cursor_getCommentRange
-# clang_Cursor_getRawCommentText
-# clang_Cursor_getBriefCommentText
+getRawCommentText(c::Union{CXCursor, CLCursor}) = convert(String, clang_Cursor_getRawCommentText(c))
+getBriefCommentText(c::Union{CXCursor, CLCursor}) = convert(String, clang_Cursor_getBriefCommentText(c))
```
Gnimuc, line 406: So, it's better to implement these two functions naively as:

```
function getRawCommentText(c::Union{CXCursor, CLCursor})
    cxstr = clang_Cursor_getRawCommentText(c)
    ptr = clang_getCString(cxstr)
    ptr == C_NULL && return ""
    s = unsafe_string(ptr)
    clang_disposeString(cxstr)
    return s
end
```

or define and reuse a `_get_cstr` helper function for internal use.

Gnimuc (changes requested):

Gnimuc:
libclang has other high-level APIs for dumping comments/documentations: https://github.com/JuliaInterop/Clang.jl/blob/d186b2d4eb72830c1b190cb5a5248421bc3ee78a/src/LibClang.jl#L2482-L2651

It will be great if all of these functionalities can be exposed by Clang.jl.

src/cursor.jl
melonedo: Thanks for pointing this out! I will fix it.

gen/generator.toml
```diff
@@ -112,6 +112,10 @@ auto_mutability_blacklist = []
 # opposite to `auto_mutability_blacklist` and has a higher priority
 auto_mutability_whitelist = []
 
+# if true, C comments are also extracted and printed as docstring.
+# By default, Clang only parses doxygen comment, pass `-fparse-all-comments` to Clang to parse non-doxygen comments.
+extract_c_comment = false
```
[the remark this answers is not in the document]
melonedo: Thanks for pointing this out! I will fix it.

melonedo:
> `libclang` has other high-level APIs for dumping comments/documentations at a fine-grained level

I have also seen this, but only getting doxygen comments as an AST is not that useful in my opinion. Maybe I need to transform that into Julia style documentation.

Gnimuc:
Yeah. GDAL.jl has [some good examples](https://github.com/JuliaGeo/GDAL.jl/blob/master/src/cpl_minixml.jl). IIRC, it first dumped the comment to a html file and then parsed them into Julia docstrings instead of using libclang's API.

Gnimuc:
@visr may chime in to give some advices.

visr:
Ah, this looks interesting!

vchuravy:
This would be really nice for LLVM.jl (cc: @maleadt)

src/cltypes.jl
```diff
@@ -73,3 +73,28 @@ Base.unsafe_convert(::Type{CXToken}, x::CLToken) = x.token
...
+    @eval begin
+        struct $clsym <: CLComment
+            comment::CXComment
+        end
+        CLCommentMap[$sym] = $clsym
+    end
+end
+
+CLComment(c::CXComment) = CLCommentMap[kind(c)](c)
+Base.convert(::Type{T}, x::CXComment) where {T<:CLComment} = CLComment(x)
```
Gnimuc, line 93:
```suggestion
Base.convert(::Type{T}, x::CXComment) where {T<:CLComment} = T(x)
```

Gnimuc (changes requested):
I just gave it a test against the LibClang itself. In general, this is great, but still has a few defects:

1. There is a leading space at the beginning of each paragraph:
```
""" Object encapsulating information about overlaying virtual file/directories over the real file system."""
const CXVirtualFileOverlay = Ptr{CXVirtualFileOverlayImpl}

"""
    CXString

 A character string.

 The `CXString` type is used to return strings from the interface when the ownership of that string might differ from one call to the next. Use `clang_getCString()` to retrieve the string data and, once finished with the string data, call `clang_disposeString()` to free the string.
"""
mutable struct CXString
    data::Ptr{Cvoid}
    private_flags::Cuint
    CXString() = new()
end
```
2. For unknown reasons, some docstrings are printed twice: `""" Describes a version number of the form major.minor.subminor."""`
```
"""
    CXVersion

 Describes a version number of the form major.minor.subminor.
"""
struct CXVersion
    """ The major version number, e.g., the '10' in '10.7.3'. A negative value indicates that there is no version number at all."""
    Major::Cint
    """ The minor version number, e.g., the '7' in '10.7.3'. This value will be negative if no minor version number was provided, e.g., for version '10'."""
    Minor::Cint
    """ The subminor version number, e.g., the '3' in '10.7.3'. This value will be negative if no minor or subminor version number was provided, e.g., in version '10' or '10.7'."""
    Subminor::Cint
end

""" Describes a version number of the form major.minor.subminor."""
"""
    CXCursor_ExceptionSpecificationKind

 Describes the exception specification of a cursor.

 A negative value indicates that the cursor is not a function declaration.
"""
@cenum CXCursor_ExceptionSpecificationKind::UInt32 begin
```

3. Both `@Cenum` and `@enum` have no support for field-docstrings, so we may need to emit everything to the docstrings of the enum itself.
```
julia> @enum Foo begin
         """bar"""
         bar,
         """baz"""
         baz
       end
ERROR: LoadError: ArgumentError: invalid argument for Enum Foo: #= REPL[5]:2 =# Core.@doc "bar" (bar, "baz")
Stacktrace:
 [1] var"@enum"(__source__::LineNumberNode, __module__::Module, T::Union{Expr, Symbol}, syms::Vararg{Any})
   @ Base.Enums ./Enums.jl:165
in expression starting at REPL[5]:1
```

4. As for struct-field-docstrings, I think we can add a style option to support 3 modes:
  * `disable`: do not emit any docstrings for fields
  * `clean`: emit all field-docstrings to a `Fields` section in the docstrings of the struct itself(like the `Arguments` section in function docstrings)
  * `default`: current style

Gnimuc:
5. The position of back-ticks is wrong or missing in some cases: 
```
"""
    CXErrorCode

 Error codes returned by libclang routines.

 Zero (`CXError_Success)` is the only error code indicating success.  Other error codes, including not yet assigned non-zero values, indicate errors.
"""

"""
    clang_File_tryGetRealPathName(file)

 Returns the real path name of `file.`

 An empty string may be returned. Use `clang_getFileName()` in that case.
"""
function clang_File_tryGetRealPathName(file)
    @ccall libclang.clang_File_tryGetRealPathName(file::CXFile)::CXString
end

"""
    CXSourceLocation

 Identifies a specific source location within a translation unit.

 Use clang_getExpansionLocation() or clang_getSpellingLocation() to map a source location to a particular file, line, and column.
"""
mutable struct CXSourceLocation
    ptr_data::NTuple{2, Ptr{Cvoid}}
    int_data::Cuint
    CXSourceLocation() = new()
end
```

Also, it would be perfect if references can be added to the code block:

```
Create a `CXVirtualFileOverlay` object. Must be disposed with `clang_VirtualFileOverlay_dispose().`
```
=>
```
Create a [`CXVirtualFileOverlay`](@ref) object. Must be disposed with [`clang_VirtualFileOverlay_dispose()`](@ref).
```

6. `\"` etc. should be stripped in the docstrings. 
```
"""
    clang_VirtualFileOverlay_addFileMapping(arg1, virtualPath, realPath)

... The virtual path must be canonicalized (not contain \".\"/\"..\").
"""
```
In fact, it doesn't matter whether we strip it or not, but the code looks a bit cleaner if we do so.

melonedo:
Obviously libclang's doxygen API is not as great as desired, so I will need some post-processing to make it feel better.

>     1. There is a leading space at the beginning of each paragraph:
Definitely libclang's issue, but it seems that in both REPL and Documenter.jl it is stripped, so this only affects source code browsing. Since a leading space is so prevalent, I can strip it for each paragraph.

>     2. For unknown reasons, some docstrings are printed twice: `""" Describes a version number of the form major.minor.subminor."""`
These are all documentation of first member of an enum. Maybe this is more desirable than no documentation if used in an interactive environment, but not if used in documentation.

>     3. Both `@cenum` and `@enum` have no support for field-docstrings, so we may need to emit everything to the docstrings of the enum itself.

>     4. As for struct-field-docstrings, I think we can add a style option to support 3 modes:
I will fix these.

>     5. The position of back-ticks is wrong or missing in some cases:
As for `CXError_Success)` It seems that libclang does not handle this correctly. Some post-processing can help.
As for missing back-ticks, `clang_getExpansionLocation` is not quoted in C source. Maybe doxygen is intelligent enough to identify these symbols.

>     6. \ in \", \\, etc. should be stripped in the docstrings.
Do you mean  `raw"""..."""`? Otherwise `\` must be escaped.

melonedo:
> The position of back-ticks is wrong or missing in some cases:
Reference: [Automatic link generation](https://www.doxygen.nl/manual/autolink.html)
One reason is that doxygen generates links based on text content instead of special commands. For example, "TypeName"(with at least one capital letter) and "function_call()"(with a pair of parentheses) are recognized as automatic links if they represent documented symbols.
We will need a set of all used symbols to automatically generated links, is this already available in Clang.jl?

Gnimuc:
> Do you mean raw"""..."""? Otherwise \ must be escaped.

Yeah. For example:

```
julia> """
       "foo"
       """
       function dummy end
dummy

help?> dummy
search: dummy MethodSummary

  "foo"

julia> """
       \"foo\"
       """
       function dummy2 end
dummy2

help?> dummy2
search: dummy2 dummy MethodSummary

  "foo"
```

> We will need a set of all used symbols to automatically generated links, is this already available in Clang.jl?

I don't quite understand the question here. If the back-ticks can be emitted correctly, we can use the same method to add `[...](@ref)` around the back-ticks. I was thinking about something like:

```
getText -> "TypeName"/"function_call()" -> extractSymbol -> "TypeName"/"function_call" -> "[`" * symbol * "`](@ref)"
```

melonedo:
> I don't quite understand the question here. If the back-ticks can be emitted correctly, we can use the same method to add [...](@ref) around the back-ticks.

Doxygen links are just plain texts, so we must filter C identifiers out of English words. For example, `CIndex` is a class name, then all appearances of "CIndex" are rendered as a link, except escaped explicitly. In your example
```c
/**
 * Identifies a specific source location within a translation
 * unit.
 *
 * Use clang_getExpansionLocation() or clang_getSpellingLocation()
 * to map a source location to a particular file, line, and column.
 */
typedef struct {
  const void *ptr_data[2];
  unsigned int_data;
} CXSourceLocation;
```
`clang_getExpansionLocation` will be rendered as a link in doxygen, although it is not annotated with `\c`. This is because doxygen assumes "clang_getExpansionLocation()" to be a function name, and it is documented, so a link is generated. For a counter-example, "Identifies" has a capital letter, but it is not a name of a class, so it is not rendered as a link.

So if we are going to support this feature, we need to test every word against the list of documented identifiers.

Gnimuc:
> `clang_getExpansionLocation` will be rendered as a link in doxygen, although it is not annotated with `\c`.

How hard is it to copy doxygen's default behavior?  For everything tailing with a `()`, we treat it as a function and emit `[`symbol`](@ref)`. It doesn't matter whether the referenced symbol has docs or not because it's `Documenter.jl`'s responsibility to handle that for us.

> For a counter-example, "Identifies" has a capital letter, but it is not a name of a class, so it is not rendered as a link.

If the design choice is to support annotated symbols only, then this is no longer a problem.

By design, we support:
1. all annotated functions/classes/etc 
2. non-annotated functions with a pattern `f()`

melonedo:
@Gnimuc  all the points you mentioned in the comments have been implemented, except
* links are all generated if they are symbols tracked by Clang.jl.
* single `"` is not escaped, but `\` is still escaped.

Gnimuc:
great! I'm going to give it a test on LLVM.jl or GDAL.jl this weekend, maybe earlier if I could find some time.

gen/generator.toml
```diff
@@ -112,6 +112,24 @@ auto_mutability_blacklist = []
 # opposite to `auto_mutability_blacklist` and has a higher priority
 auto_mutability_whitelist = []
 
+# if set to "raw", extract and dump raw c comment;
+# if set to "doxygen", parse and format doxygen comment.
+# note: by default, Clang only parses doxygen comment, pass `-fparse-all-comments` to Clang in order to parse non-doxygen comments.
+extract_c_comment_style = "disable"
+
+# if set to true, single line comment will be printed as """comment""" instead of """\ncomment\n"""
+fold_single_line_comment = true
```
Gnimuc, line 121:
```suggestion
fold_single_line_comment = false
```

to match the default behavior in the code: `fold_single_line_comment = get(options, "fold_single_line_comment", false)`

Gnimuc (approved):
The generated docs for LLVM.jl look good. I believe this is a feature that everyone would love!

Gnimuc:
@melonedo do you have time to update GDAL.jl as requested in https://github.com/JuliaInterop/Clang.jl/pull/318#issuecomment-886242330 ?

src/generator/documentation.jl
```diff
@@ -0,0 +1,399 @@
...
+end
+
+function get_prototype(node)
+    code = Clang.getSourceCode(node.cursor)
+    code = replace(code, r"/\*.*?\*/|//.*?$"m=>"")
+    code = replace(code, r"\s+"m=>" ")
+    "$code;"
+end
+
+escape_underscore(word) = replace(word, "_"=>"\\_")
```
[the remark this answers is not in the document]
melonedo: `_` is synonym for `*` in markdown. However each paragraph is a separate markdown context, so the markdown parser can not match `_` pairs in this example, and falls back to show it verbatim. You can check `CXCursor_ExceptionSpecificationKind` where this matters:
`CXCursor_ExceptionSpecificationKind_None` will be shown as CXCursor*ExceptionSpecificationKind*None.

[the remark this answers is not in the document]
Gnimuc: That makes sense to me now. Thanks!

melonedo:
> @melonedo do you have time to update GDAL.jl as requested in [#318 (comment)](https://github.com/JuliaInterop/Clang.jl/pull/318#issuecomment-886242330) ?

Sure! I will try it today.

Gnimuc:
I think we can merge this and add new features or do bugfixes in other PR.

gen/generator.toml
melonedo: Is this file the default behavior or only what is used to generate LibClang.jl? I find that a few other options are not the default behavior, for example, "use_deterministic_symbol" defaults to false.

Gnimuc: For those options that are not currently being used for generating libclang C bindings, so better to make them consistent with the default behavior in the code. But apparently, I did it wrong in the case of `auto_mutability`...

PR #319, opened 2021-07-24
Manually throw error for incomplete expression

### LINKED ISSUE #311: "$" expression outside quote around...
I'm attempting to wrap `<sys/mman.h>` which runs successfully but generators the following line:

```julia
const __ptr_t = $(Expr(:incomplete, "incomplete: premature end of input"))
```

resulting in the error:
```
LoadError: syntax: "$" expression outside quote around /home/colinxs/workspace/crypto/obs/gen/mman.jl:313
```

The corresponding c code is in `<sys/cdefs.h>`:
```c
/* This is not a typedef so `const __ptr_t' does the right thing.  */
#define __ptr_t void *
```

I can blacklist it for now, since it's not used directly in the generated code, but figured I'd file an issue to put it on the radar.

melonedo:
Fix #311

### DIFF
--- a/src/generator/macro.jl
+++ b/src/generator/macro.jl
@@ -271,6 +271,34 @@ function macro_emit!(dag::ExprDAG, node::ExprNode{MacroUnsupported}, options::Di
     return dag
 end
 
+"Try to parse the string, return `nothing` if the string is incomplete or has syntax error."
+function try_parse(str)
+    try
+        ex = Meta.parse(str)
+        Meta.isexpr(ex, :incomplete) && throw(Meta.ParseError("failed to parse:  $str"))
+        return ex
+    catch err
+        return nothing
+    end
+end
+
+"Parse a string potentially as a pointer type (`type *``). Issue #311"
+function parse_as_pointer(str)
+    m = match(r"^(.+?)((?:\s(:?\*|const|volatile))+)$", str)
+    isnothing(m) && return nothing
+
+    base_type = m[1]
+    ptr_postfix = m[2]
+    ex = try_parse(base_type)
+    isnothing(ex) && return nothing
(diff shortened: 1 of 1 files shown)


src/generator/macro.jl
```diff
@@ -296,7 +296,9 @@ function macro_emit!(dag::ExprDAG, node::ExprNode{MacroDefault}, options::Dict)
     end
     str = reduce(add_spaces_for_macros, txts)
     try
-        push!(node.exprs, Expr(:const, Expr(:(=), sym, Meta.parse(str))))
+        ex = Meta.parse(str)
+        Meta.isexpr(ex, :incomplete) && throw(ParseError(ex.args|>only))
```
Gnimuc, line 300:
```suggestion
        Meta.isexpr(ex, :incomplete) && throw(Meta.ParseError("failed to parse:  $str"))
```

Gnimuc (commented):
To fix #311, we should add support for this macro pattern instead of skipping it.

melonedo:
Can we support this kind of macro? `void *` is tricky even in C...
But at least we must ensure the generated code is valid Julia code, so I forced the generator to ignore this macro.

Gnimuc:
No need to be generic. In this specific case, we can check: if the parsed expr is `incomplete` and the last char in `str` is `*`, then we remove this `*` from `str` and parse again until all of the tailing `*` are recursively removed; now the only thing left should be a valid Julia type, if not, we just give up and skip.

Gnimuc:
@melonedo, is this ready to merge?

melonedo:
I think yes. I do not see any regression locally.

Gnimuc:
Thanks!

PR #321, opened 2021-08-09
Make TypedefElaborated respect `@add_def`

### LINKED ISSUE #320: Unable to wrap `stat`
This happened when I am wrapping GDAL.jl.
On linux, if I attemp to wrap this code:
```c
#include <sys/stat.h>
typedef struct stat mystat;
```
because `stat` is also a function, `Audit` pass can not be passed. But it seems that neither `CollectTopLevelNode` nor `CollectDependentSystemNode` supports filtering, and `Audit` pass can not be disabled.
Do we need to add filtering earlier to solve this?

melonedo:
Fix #320

### DIFF
--- a/src/generator/definitions.jl
+++ b/src/generator/definitions.jl
@@ -12,7 +12,6 @@ const DEFAULT_EXTRA_DEFINITIONS = Dict{Symbol,AbstractJuliaType}(
     :int16_t => JuliaCint16_t(),
     :int8_t => JuliaCint8_t(),
     :uintptr_t => JuliaCuintptr_t(),
-    :tm => JuliaCtm(),
     :wchar_t => JuliaCwchar_t(),
     :va_list => JuliaUnsupported(),
     :FILE => JuliaCFILE(),
--- a/src/generator/resolve_deps.jl
+++ b/src/generator/resolve_deps.jl
@@ -87,6 +87,8 @@ function resolve_dependency!(dag::ExprDAG, node::ExprNode{TypedefElaborated}, op
 
     if haskey(dag.tags, leaf_ty.sym)
         push!(node.adj, dag.tags[leaf_ty.sym])
+    elseif haskey(dag.ids_extra, leaf_ty.sym)
+        # pass
     else
         tycu = getTypeDeclaration(ty)
         file, line, col = get_file_line_column(cursor)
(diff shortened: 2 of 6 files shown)


Gnimuc:
[This line](https://github.com/JuliaInterop/Clang.jl/blob/master/src/generator/definitions.jl#L15) can be removed as well.

Gnimuc:
The `@add_def` macro was originally designed for adding identifiers in the system headers. After adding multi-platform support, this macro is mainly for skipping something wrong or unexpected. As a result, the current implementation is a mess and needs to be refactored in the future.

melonedo:
> [This line](https://github.com/JuliaInterop/Clang.jl/blob/master/src/generator/definitions.jl#L15) can be removed as well.

Is this because currently tags are not translated?

Gnimuc:
As mentioned above, initially `@add_def` was for adding identifiers in the system headers, at that time, all the system-header definitions need to be manually wrapped. 

Then, when implementing the multi-platform support, `@add_def` was for extracting specific identifiers in the system headers, for example, `@add_def tm` can be used to tell the generator to search `tm` in the system headers and generate a wrapper for `tm`. This was still very cumbersome because users need to add those `@add_def`s manually. 

Then, the experimental [`CollectDependentSystemNode ` pass](https://github.com/JuliaInterop/Clang.jl/blob/master/src/generator/passes.jl#L61-L91) was added, so all dependent system nodes can be automatically handled. So, `@add_def` is now only for doing workaround. 

This PR adds `|| haskey(dag.ids_extra, leaf_ty.sym))` for `TypedefElaborated` to make the behavior consistent with `TypedefDefault`, which is mainly for fixing the old behavior for extracting typedef identifiers in the system headers.
However, [this line](https://github.com/JuliaInterop/Clang.jl/blob/master/src/generator/definitions.jl#L15) was for rewriting the definition of `tm` to a custom one, which is a misuse.

For now, the `@add_def` macro has two different meanings for typedef-ed identifiers and tag types. It needs to be separated into two macros, one for extracting identifiers `@extract_def` and one for skipping tag types `@skip_def`. But it's not trivial to do the refactor.

melonedo:
I added a test case for this.

melonedo:
The tests failures are with Julia 1.7. New CI scripts is not enabled.

Gnimuc:
need to rebase this PR onto master.

melonedo:
Oops this keeps having conflicts as new tests are added.

PR #323, opened 2021-08-16
`Char_U` should be mapped to `JuliaCuchar`

Gnimuc:
Maybe it's better to add a new option to let users determine the char type they'd like to use on the Julia side.


Gnimuc:
turns out that it actually doesn't make much difference when doing Julia-C interop.

PR #330, opened 2021-09-09
Update freebsd

melonedo:
Since JuliaPackaging/BinaryBuilderBase.jl#166 has removed support for freebsd 11 and switched to freebsd 12, this PR updates freebsd version, and renames freebsd11.1 as generic freebsd12.2.


Gnimuc:
LGTM.

PR #331, opened 2021-09-09
Resolve dependency for unnamed struct and macro

### LINKED ISSUE #325: Wrong order of generated structs for generation of structs containing anonymous unions
Hi everyone, 
thank you for the amazing package so far!

I think I encountered a small (but annoying) bug in the current master branch: the generation of structs containing anonymous unions prints the corresponding Julia union wrapper struct after the original struct definition. 

In my case, the C struct

```C  
typedef struct {
      enum NodeIdType id_type;
      union {
          UInt32     numeric;
          Guid       guid;
      } id;
  } NodeId;
```
results in the generated code

```julia
mutable struct NodeId
    id_type::NodeIdType
    id::__JL_Ctag_295
    NodeId() = new()
end
...
struct __JL_Ctag_295
    data::NTuple{16, UInt8}
end
```
leading to 
```
ERROR: LoadError: LoadError: UndefVarError: __JL_Ctag_295 not defined
```
If I am not missing something, the problem can be solved by changing the order of the generated structs.

melonedo:
* Resolve dependency for unnamed struct.
* `resolve_dependency!` now accepts options as last parameter.
* Resolve dependency for macros. This commit is also included because of the API change above.
Fix #325


melonedo:
I think I can add a few test cases for this.

PR #335, opened 2021-09-13
Support dumping enum as integer constants

melonedo:
This PR reuses the old path that emits enum as an alias for integer type.
The second commit made enum carry necessary information for further analysis, but my trial to use it did not succeed. Should I keep this?


Gnimuc:
> The second commit made enum carry necessary information for further analysis, but my trial to use it did not succeed. Should I keep this?

That looks good to me.

PR #338, opened 2021-09-16
Revising generator.md

mkitti:
This patch is just a minor editing of the generator.md text after proofreading.


Gnimuc:
Thanks mkitti!

PR #341, opened 2021-09-22
Handle typedef to pointer of anonymous struct

### LINKED ISSUE #326: Can not handle pointer to anonymous struct
```c
typedef struct {
    int a, b;
} B, *C;
```
Clang.jl can not distinguish B from C.

melonedo:
Fix #326.


Gnimuc:
Thanks!

PR #346, opened 2021-10-13
Revert "Merge pull request #302 from logankilpatrick/master"

KristofferC:
This reverts commit bacacb4ac16f7dccefc489d9e3af5602d8422c7b, reversing
changes made to 104c792b50ad2175adf44a52dc28ec12d7558032.

`denylist` is not a good replacement for the existing term because nothing is being "denied". `exclusions` or something along those lines would be more suitable.

`allowlist` is also bad, `inclusion` something would be better.

### DIFF
--- a/gen/generator.toml
+++ b/gen/generator.toml
@@ -44,9 +44,9 @@ prologue_file_path = "./prologue.jl"
 # this is often used for applying custom patches.
 epilogue_file_path = ""
 
-# node with an id in the `printer_denylist` will be ignored in the printing passes. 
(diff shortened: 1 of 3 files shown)


KristofferC:
> seems to be the official replacement for whitelist/blacklist,

There aren't really any official replacements. It is all about finding new words that somewhat describe as well as possible the intent of the original words. When one would describe the usage here in a sentence you would typically say that you would "include" or "ignore" some ids. You would never say that you would "deny" the ids, that doesn't make sense. So using something like `include_list` and `exclude_list` might make sense?

Issue #344, opened 2021-10-13
Add a mode to only generate a necessary set of API


Gnimuc:
Clang.jl's generator is mainly for creating a complete self-contained Julia binding wrapper package for C libraries. 

For C libraries with a huge API set, it's easy to generate raw bindings with Clang.jl but hard to maintain a set of high-level Julia wrappers. So, in practice, there are cases where users only want to use, maintain or embed a small group of API functions from a large C library in their projects. 

This problem can be alleviated by adding a "lazy" mode to Clang.jl's generator. Given a set of names, this mode should recursively generate bindings for functions whose names are matched in the set and all dependent identifiers.

PR #360, opened 2021-12-20
add an exclusivity list to output options

KristofferC:
If this list is given, only symbols in the list will be output.

If this is deemed an ok idea, I can add tests etc.

### DIFF
--- a/gen/generator.toml
+++ b/gen/generator.toml
@@ -58,9 +58,6 @@ output_ignorelist = [
     "LLVM_CLANG_C_EXTERN_C_END"
 ]
 
-# Only output id in the `output_exclusivelist"
-# output_output_exclusivelist = []
-
 # Julia's `@enum` do not allow duplicated values, so by default, C enums are translated to 
 # CEnum.jl's `@cenum`. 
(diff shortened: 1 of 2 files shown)


Gnimuc:
This seems to be the first step to #344

KristofferC:
So this should also get everything dependent on the symbols in the exclusive list?

Gnimuc:
> If this list is given, only symbols in the list will be output.

If a symbol in this list depends on other symbols that are not in this list, then the generated Julia code can not be run. 

What #344 described is that we automatically create a minimal symbol set from this list to make sure the generated code is runable.

gen/generator.toml
```diff
@@ -58,6 +58,9 @@ output_ignorelist = [
     "LLVM_CLANG_C_EXTERN_C_END"
 ]
 
+# Only output id in the `output_exclusivelist"
+# output_output_exclusivelist = []
+
```
Gnimuc, line 63:
```suggestion
```
I think we can merge this and keep this feature private for now.

KristofferC:
Thanks! I have use for this in a wrapper I am toying with.

PR #365, opened 2022-01-02
3 minor NFC changes

visr:
Unrelated but in 3 separate commits. I thought I'd bundle them since they are probably rather straightforward.


Gnimuc:
Thanks!

PR #366, opened 2022-01-02
strip trailing whitespace in generated docstrings

visr:
And also remove remove empty lines in Parameters or Returns lists. Though it could be that I don't understand the reason those need to be there. The markdown format doesn't require them.

- remove trailing whitespace for Clang.Paragraph
see the effect in https://github.com/JuliaGeo/Proj4.jl/commit/04b9fb8136ea1821bca8c370de59e4bfeabffe9b
- remove empty lines in Parameters or Returns lists
see the effect in https://github.com/JuliaGeo/Proj4.jl/commit/fc126ae7e6f7a3b2fb51c96eba15c7d189054847
- strip trailing whitespace after parameter
see the effect in https://github.com/JuliaGeo/Proj4.jl/commit/7830d54720c3154e3062a7e6841af35a3256290d

### DIFF
--- a/src/generator/documentation.jl
+++ b/src/generator/documentation.jl
@@ -243,13 +243,11 @@ function format_doxygen(cursor, options, members=false)
         push!(lines, "### Parameters")
         for p in parameters
             append!(lines, format_parameter(p, options))
-            push!(lines, "")
         end
     end
     if !ismissing(returns)
         push!(lines, "### Returns")
         append!(lines, format_block(Clang.getParagraph(returns), options))
-        push!(lines, "")
     end
     if !isempty(member_docs)
         append!(lines, member_docs)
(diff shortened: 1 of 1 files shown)


melonedo:
From [Julia manual on markdown lists](https://docs.julialang.org/en/v1/stdlib/Markdown/#Lists), the empty line is required for complex markdown elements. For example a nested list in parameter.
```md
## Parameters
- `arg`: This arguments takes various forms:
    - form1
    - form2
- `arg2`: code block

    ```c
    int foo(void);
    ````

```
My experiments showed no difference though.

Gnimuc:
That doc is probably outdated. [This change](https://github.com/JuliaLang/julia/pull/26598) might be related.

melonedo:
So we can omit the empty line as long as we do not find broken format. I could not reproduce what formatting issue might be triggered by omitting that.

PR #367, opened 2022-01-04
Add functinal tests for bitfield

melonedo:
I built a simple C library to test against Clang.jl's bitfield implementation. The results seem fine.

### DIFF
--- a/test/bitfield/build_tarballs.jl
+++ b/test/bitfield/build_tarballs.jl
@@ -1,0 +1,35 @@
+# Note that this script can accept some limited command-line arguments, run
+# `julia build_tarballs.jl --help` to see a usage message.
+using BinaryBuilder, Libdl
+
+name = "libbitfield"
+version = v"0.0.1"
+
+# Collection of sources required to complete build
+sources = [
+    DirectorySource(@__DIR__),
+]
+
+# Bash recipe for building across all platforms
+script = raw"""
+cd $WORKSPACE/srcdir
+cmake -B build -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DCMAKE_BUILD_TYPE=Release
+cmake --build build
+cmake --install build
+touch LICENSE
+"""
+
+# Only need to build for host platform since we are not publishing
+platforms = [HostPlatform()]
+
+# The products that we will ensure are always built
+products = [
+    LibraryProduct(["libbitfield"], :libbitfield)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
--- a/test/test_bitfield.jl
+++ b/test/test_bitfield.jl
@@ -8,10 +8,14 @@ function build_libbitfield()
 
     config_cmd = `$cmake -B $build_dir -S $src_dir`
     if Sys.WORD_SIZE == 32
-        Sys.iswindows() && push!(config_cmd.exec, "-A", "win32")
-        # Not actually working, so fail fast
-        Sys.islinux() && error("Can not build libbitfield on 32-bit linux")
-        Sys.islinux() && push!(config_cmd.exec, "-D", "CMAKE_C_FLAGS=-m32", "-D", "CMAKE_CXX_FLAGS=-m32")
+        if Sys.iswindows()
+            config_cmd = `$config_cmd -A win32`
+        elseif Sys.islinux()
+            config_cmd = `$config_cmd -D CMAKE_C_FLAGS=-m32 -D CMAKE_CXX_FLAGS=-m32`
+            # Not actually working, so fail fast
+            @info "Can not build libbitfield on 32-bit linux"
+            error()
+        end
     end
     build_cmd = `$cmake --build $build_dir --config Debug`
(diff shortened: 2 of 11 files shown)


test/test_bitfield.jl
```diff
@@ -0,0 +1,70 @@
...
+    build_dir = joinpath(@__DIR__, "build", "bitfield")
+
+    config_cmd = `$cmake -B $build_dir -S $src_dir`
+    if Sys.WORD_SIZE == 32
+        if Sys.iswindows()
+            config_cmd = `$config_cmd -A win32`
+        elseif Sys.islinux()
+            config_cmd = `$config_cmd -D CMAKE_C_FLAGS=-m32 -D CMAKE_CXX_FLAGS=-m32`
+            # Not actually working, so fail fast
+            @info "Can not build libbitfield on 32-bit linux"
```
Gnimuc, line 16: Just out of curiosity. Why does this test not love 32-bit Linux?

melonedo:
Without these flags, CI builds x64 on x86; with these flags CI can not even compile anything. So I disable it until someone figures out how to do that.



---Original---
From: "Yupei ***@***.***&gt;
Date: Tue, Jan 4, 2022 17:49 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [JuliaInterop/Clang.jl] Add functinal tests for bitfield (PR #367)




 
@Gnimuc commented on this pull request.
 
 
In test/test_bitfield.jl:
 &gt; +using CMake +using Clang.Generators + +function build_libbitfield() +    @info "Building libbitfield. This will break if CMake fails to find a C compiler" +    src_dir = joinpath(@__DIR__, "bitfield") +    build_dir = joinpath(@__DIR__, "build", "bitfield") + +    config_cmd = `$cmake -B $build_dir -S $src_dir` +    if Sys.WORD_SIZE == 32 +        if Sys.iswindows() +            config_cmd = `$config_cmd -A win32` +        elseif Sys.islinux() +            config_cmd = `$config_cmd -D CMAKE_C_FLAGS=-m32 -D CMAKE_CXX_FLAGS=-m32` +            # Not actually working, so fail fast +            @info "Can not build libbitfield on 32-bit linux"  
Just out of curiosity. Why does this test not love 32-bit Linux?
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

Gnimuc:
An alternative way is to use BinayBuilder.jl for building that little library.

melonedo:
BinaryBuilder does not offer a easy solution for building on current platform only, and I think downloading toolkit and building for every platform seems like too much overhead on a local test machine.

test/test_bitfield.jl
```diff
@@ -0,0 +1,67 @@
...
+        build_dir = joinpath(@__DIR__, "build", "bitfield")
+
+        config_cmd = `$cmake -B $build_dir -S $src_dir`
+        if Sys.WORD_SIZE == 32
+            if Sys.iswindows()
+                config_cmd = `$config_cmd -A win32`
+            elseif Sys.islinux()
+                config_cmd = `$config_cmd -D CMAKE_C_FLAGS=-march=native -D CMAKE_CXX_FLAGS=-march=native`
+            end
+        end
```
Gnimuc, line 19: I'm wondering why CMake can not handle this for us.

melonedo:
These flags are useless in the sense that whether with these flags or not CMake will fail because "gcc can not compile a trivial program" with the combination of flags.
Maybe the linux-32bit mirror does not have a working gcc for its own architecture.



---Original---
From: "Yupei ***@***.***&gt;
Date: Mon, Jan 10, 2022 13:54 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [JuliaInterop/Clang.jl] Add functinal tests for bitfield (PR#367)




 
@Gnimuc commented on this pull request.
 
 
In test/test_bitfield.jl:
 &gt; +        if Sys.WORD_SIZE == 32 +            if Sys.iswindows() +                config_cmd = `$config_cmd -A win32` +            elseif Sys.islinux() +                config_cmd = `$config_cmd -D CMAKE_C_FLAGS=-march=native -D CMAKE_CXX_FLAGS=-march=native` +            end +        end  
I'm wondering why CMake can not handle this for us.
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

melonedo:
It turns out that BinaryBuilder is less accessible than native tool though, it only works on ubuntu-x64 platform.

test/bitfield/build_tarballs.jl
```diff
@@ -0,0 +1,35 @@
...
+products = [
+    LibraryProduct(["libbitfield"], :libbitfield)
+]
+
+# Dependencies that must be installed before this package can be built
+dependencies = Dependency[
+]
+
+# Build the tarballs, and possibly a `build.jl` as well.
+build_tarballs(ARGS, name, version, sources, script, platforms, products, dependencies)
```
Gnimuc, line 35: BinaryBuilder is not supposed to be used in this way. It's for building and deploying binaries in a controlled environment. After building the binary, you need to deploy it somewhere by running `julia build_tarballs.jl --deploy=.....`, and then use the deployed JLL package for testing. 

As this is a library merely for testing Clang.jl, so there is no need to use Yggdrasil for deploying.

Gnimuc:
I'm merging this as it may not be worth the hassle. 32bit systems will quickly fade into history.

Issue #317, opened 2021-07-22
No tests for wrapper correctness


melonedo:
Currently, we only have tests for the integrity of the generated wrapper, while correctness needs to be tested on downstream packages. 
Maybe we could build a custom test library in C (with BinaryBuilder), then relevant tests can be hosted on Clang.jl itself.

Gnimuc:
+1 for this. 

[There were some test cases before](https://github.com/JuliaInterop/Clang.jl/tree/5971591cd4bf0d8b128da3d8cb7a2649f6c2e665/test). I removed them because all of these test cases were too trivial.

For new features like bitfield, we do need decent unit tests. It's not easy to work with BB2 when developing the package interactively. It turns out this is probably an ideal use case for [`ClangComplier.jl`](https://github.com/Gnimuc/ClangComplier.jl) with which we can compile the C code within Julia and call the library through LLVM.jl.

JamesWrigley:
Closing because we now have correctness tests for bitfields. We can extend that later if needed.

PR #368, opened 2022-01-09
Generators: Add padding between non consecutive fields

Pangoraw:
Hello :wave:

I had an issue with Clang.jl where the created Julia struct would not match the C representation of the same struct because of field alignments issues.

Consider the following struct:

```c
struct B {
  uint64_t ba;
  uint64_t bb;
};

union union_B {
  struct B b;
};

struct A {
  char a;
  union union_B b;
};

/**/
sizeof(struct A) // 24
sizeof(struct B) // 16
sizeof(struct A) != sizeof(char) + sizeof(struct B) // true
```

The generated Julia struct is:

```julia
struct B
  ba::UInt64
  bb::UInt64
end

struct union_B
    data::NTuple{16, UInt8}
end

struct A
  a::Cchar
  b::union_B
end

#==#
sizeof(A) # 17
sizeof(union_B) # 16
sizeof(A) != sizeof(Cchar) + sizeof(union_B) # false
```

The C and Julia version of field `b::B` are not aligned (C is `+8` whereas Julia is only `+1`). This PR tries to address this problem by adding new fields on the Julia side and adding a custom constructor to the struct:

```julia
struct B
  ba::UInt64
  bb::UInt64
end

struct union_B
    data::NTuple{16, UInt8}
end

struct A
    a::Cchar
    var"##pad0#274"::NTuple{7, UInt8}
    b::union_B
    A(a, b) = new(a, Tuple((zero(UInt8) for _ = 1:7)), b)
end

#==#
sizeof(A) # 24
sizeof(union_B) # 16
sizeof(A) != sizeof(Cchar) + sizeof(union_B) # true
```

I have noticed that there are only a few tests for `Clang.Generators` so I would be interested on having other codebase to test this code on. For example when generating the Julia for `clang-c` into LibClang some struct are added padding fields. Also, should I add test for this ? It would eval the generated julia code and check the different `sizeof`.

Anyway, thanks for making this project!

### DIFF
--- a/src/generator/codegen.jl
+++ b/src/generator/codegen.jl
@@ -482,11 +482,44 @@ function emit!(dag::ExprDAG, node::ExprNode{<:AbstractStructNodeType}, options::
     expr = Expr(:struct, false, struct_sym, block)
     field_cursors = fields(getCursorType(node.cursor))
     field_cursors = isempty(field_cursors) ? children(node.cursor) : field_cursors
-    for field_cursor in field_cursors
+
+    current_offset = 0
+    pads = []
+    for (i, field_cursor) in enumerate(field_cursors)
         field_sym = make_symbol_safe(name(field_cursor))
         field_ty = getCursorType(field_cursor)
+        field_offset = getOffsetOfField(field_cursor) ÷ 8
+        field_size = getSizeOf(field_ty)
+        @assert field_offset >= current_offset
+
+        if field_offset != current_offset
+            pad_size = field_offset - current_offset
+            pad_name = Symbol("pad", length(pads)) |> gensym
+            push!(block.args, Expr(:(::), pad_name, :(NTuple{$pad_size, UInt8})))
+            push!(pads, (;narg = i + length(pads), size = pad_size))
+        end
+
+        current_offset += field_size
         push!(block.args, Expr(:(::), field_sym, translate(tojulia(field_ty), options)))
     end
+
+    if length(pads) != 0 # build custom constructor with padding fields set to 0
+        lhs = Expr(:call, struct_sym)
+        rhs = Expr(:call, :new)
+        pads = reverse(pads)
+        for (i, ex) in enumerate(block.args)
+            maybe_pad = length(pads) > 0 && pads[end].narg == i ? pop!(pads) : nothing
+            if maybe_pad === nothing
+                field_name = first(ex.args)
+                push!(lhs.args, field_name)
+                push!(rhs.args, field_name)
+            else
+                push!(rhs.args, :(Tuple(zero(UInt8) for _ in 1:$(maybe_pad.size))))
+            end
+        end
+        push!(block.args, Expr(:(=), lhs, rhs))
+    end
+
     push!(node.exprs, expr)
 
     if startswith(string(node.id), "##Ctag") || startswith(string(node.id), "__JL_Ctag")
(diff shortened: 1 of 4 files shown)


Gnimuc:
Could you share some system info? I can't reproduce the wrong padding result on the Julia side.

```
julia> ctx = create_context("test.h", args);
[ Info: Parsing headers...

julia> build!(ctx)
[ Info: Processing header: test.h
[ Info: Building the DAG...
└ @ Clang.Generators ~/.julia/dev/Clang/src/generator/audit.jl:16
[ Info: Emit Julia expressions...
struct B
    ba::UInt64
    bb::UInt64
end

struct A
    a::Cchar
    b::B
end

[ Info: Done!
Context(...)

julia> struct B
           ba::UInt64
           bb::UInt64
       end

julia> struct A
           a::Cchar
           b::B
       end

julia> sizeof(A)
24

julia> sizeof(B)
16
```

Gnimuc:
> I have noticed that there are only a few tests for `Clang.Generators` so I would be interested on having other codebase to test this code on.

It would be great if you'd like to contribute a set of test cases. There were some simple tests as mentioned in https://github.com/JuliaInterop/Clang.jl/issues/317#issuecomment-884837470 and a [repo](https://github.com/Gnimuc/GeneratorScripts) for monitoring the sanity of the generated code.

Pangoraw:
oh I simplified my example a bit too much. It only happens when the re-aligned field is an union because in Julia they are represented as `NTuple{X, UInt8}` which does not require a different alignment. I updated the example as such.

src/generator/codegen.jl
```diff
@@ -482,11 +482,44 @@ function emit!(dag::ExprDAG, node::ExprNode{<:AbstractStructNodeType}, options::
...
+        @assert field_offset >= current_offset
+
+        if field_offset != current_offset
+            pad_size = field_offset - current_offset
+            pad_name = Symbol("pad", length(pads)) |> gensym
+            push!(block.args, Expr(:(::), pad_name, :(NTuple{$pad_size, UInt8})))
+            push!(pads, (;narg = i + length(pads), size = pad_size))
+        end
+
+        current_offset += field_size
```
Gnimuc, line 502: This fix only handles generic cases, the same fix should also apply for `StructMutualRef`.

Gnimuc:
An alternative fix would be marking these kinda structs as `StructLayout` in the preprocessing pass like what we do for [`nested_anonymous_type`](https://github.com/JuliaInterop/Clang.jl/blob/331313d58c39d102aa0bc7b09a9d1473f16d373e/src/generator/passes.jl#L583-L588)s. In this way, we redirect its codegen pass to [here](https://github.com/JuliaInterop/Clang.jl/blob/331313d58c39d102aa0bc7b09a9d1473f16d373e/src/generator/codegen.jl#L585), so don't need to change anything in the CodeGen pass.

Gnimuc:
```
struct A
    a::Cchar
    var"##pad0#274"::NTuple{7, UInt8}
    b::union_B
    A(a, b) = new(a, Tuple((zero(UInt8) for _ = 1:7)), b)
end
```

I think this solution is also good. We can add an option in the .toml file to let users choose what they'd like to generate.

You could implement this by adding a new type like `StructLayout` and writing a new codegen pass for it.

Gnimuc:
https://github.com/JuliaInterop/Clang.jl/pull/368#issuecomment-1009602873 has been implemented in #370 and merged to the master.

Pangoraw:
> [#368 (comment)](https://github.com/JuliaInterop/Clang.jl/pull/368#issuecomment-1009602873) has been implemented in #370 and merged to the master.

Awesome! Sorry i didn't had time to update on this. I don't know whether or not it's worth adding the different codegen with paddings now that it uses a `StructLayout`. What do you think ?

PR #372, opened 2022-01-30
feat(JLLEnvs): try to load local Artifacts.toml if existed

johnnychen94:
Almost certainly I get network issues when `using Clang` in China unless I set proxies. This PR works around it so I figured people might want to see this change.

```
julia> using Clang
ERROR: InitError: Failed to connect to raw.githubusercontent.com port 443: Connection refused while requesting https://raw.githubusercontent.com/JuliaPackaging/BinaryBuilderBase.jl/master/Artifacts.toml
```

---

This commit provides two changes:

- By default, use the stable release version of `Artifacts.toml` provided by
  binarybuilderbase. previously it uses the master version, which can sometimes
  be breaking and troublesome.
- it utilizes the local disk cache and thus pkg server so that we don't need to
  fetch resources from GitHub.com and thus is more friendly to users behind the
  firewall.

As a consequence of getting rid of the need to download from remote server,
this commit makes `@time using Clang` much faster: from 1.7s to 0.4s in Julia
1.7, macOS Intel i9-9880H.

### DIFF
--- a/src/shards/JLLEnvs.jl
+++ b/src/shards/JLLEnvs.jl
@@ -3,18 +3,41 @@ module JLLEnvs
 using Pkg
 using Pkg.Artifacts
 using Downloads
+using UUIDs
 
 include("utils.jl")
 
 const JLL_ENV_SHARDS_URL = "https://raw.githubusercontent.com/JuliaPackaging/BinaryBuilderBase.jl/master/Artifacts.toml"
 const JLL_ENV_SHARDS = Dict{String,Any}()
 
+function _checked_import(pkgid)
+    Base.root_module_exists(pkgid) && return Base.root_module(pkgid)
+    Base.require(pkgid)
+end
+
 function __init__()
     if haskey(ENV, "JULIA_CLANG_SHARDS_URL") && !isempty(get(ENV, "JULIA_CLANG_SHARDS_URL", ""))
         merge!(JLL_ENV_SHARDS, Artifacts.load_artifacts_toml(ENV["JULIA_CLANG_SHARDS_URL"]))
-    else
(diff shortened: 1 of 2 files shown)


src/shards/JLLEnvs.jl
```diff
@@ -3,18 +3,41 @@ module JLLEnvs
...
+    Base.require(pkgid)
+end
+
 function __init__()
     if haskey(ENV, "JULIA_CLANG_SHARDS_URL") && !isempty(get(ENV, "JULIA_CLANG_SHARDS_URL", ""))
         merge!(JLL_ENV_SHARDS, Artifacts.load_artifacts_toml(ENV["JULIA_CLANG_SHARDS_URL"]))
-    else
-        merge!(JLL_ENV_SHARDS, Artifacts.load_artifacts_toml(Downloads.download(JLL_ENV_SHARDS_URL)))
+        return
     end
```
johnnychen94, line 22: Still keep the `JULIA_CLANG_SHARDS_URL` environment flag of the highest priority so that people can still debug with specific Artifacts.toml version.

melonedo:
This is very nice! I have always manually redirected that Artifacts.toml to a local file before. Additionally, we can also avoid directly calling the `Pkg.download_artifact` function instead of `Pkg.ensure_artifact_downloaded`, which respects pkg mirror.

johnnychen94:
> I'm considering directly shipping this Artifact.toml in Clang.jl.

I've thought about this too, but then I feel it brings more future synchronization/maintenance work to do without CompatHelper-like CI tools.

Gnimuc:
closed in favor of #377

PR #377, opened 2022-02-17
Use pkg mirror to download artifacts

melonedo:
`Artifacts.ensure_artifact_installed` will respect pkg mirror, which avoids directly downloading from GitHub, as mentioned in https://github.com/JuliaInterop/Clang.jl/pull/372#issuecomment-1025451144.

### DIFF
--- a/src/shards/JLLEnvs.jl
+++ b/src/shards/JLLEnvs.jl
@@ -8,12 +8,18 @@ include("utils.jl")
 
 const JLL_ENV_SHARDS = Dict{String,Any}()
 
+const ARTIFACT_TOML_PATH = Ref{String}()
+
+# The platform used in Artifacts.toml
+const BINARY_BUILDER_PLATFORM = Base.BinaryPlatforms.Platform("x86_64", "linux"; libc="musl")
+
 function __init__()
     if haskey(ENV, "JULIA_CLANG_SHARDS_URL") && !isempty(get(ENV, "JULIA_CLANG_SHARDS_URL", ""))
-        merge!(JLL_ENV_SHARDS, Artifacts.load_artifacts_toml(ENV["JULIA_CLANG_SHARDS_URL"]))
+        ARTIFACT_TOML_PATH[] = ENV["JULIA_CLANG_SHARDS_URL"]
     else
-        merge!(JLL_ENV_SHARDS, Artifacts.load_artifacts_toml(joinpath(@__DIR__, "..", "..", "Artifacts.toml")))
+        ARTIFACT_TOML_PATH[] = normpath(joinpath(@__DIR__, "..", "..", "Artifacts.toml"))
     end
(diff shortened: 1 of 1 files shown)


src/shards/JLLEnvs.jl
```diff
@@ -120,9 +126,9 @@ function get_system_dirs(triple::String, version::VersionNumber=v"4.8.5")
 
     # download shards
     if haskey(ENV, "JULIA_CLANG_SHARDS_URL") && !isempty(get(ENV, "JULIA_CLANG_SHARDS_URL", ""))
-        @info "Downloading artifact($(gcc_info.id)) from $(gcc_info.url) ..."
+        @info "Downloading artifact($(gcc_info.id))"
     end
-    Artifacts.download_artifact(Base.SHA1(gcc_info.id), gcc_info.url, gcc_info.chk)
+    Artifacts.ensure_artifact_installed(get_gcc_shard_key(triple, version), ARTIFACT_TOML_PATH[]; platform=BINARY_BUILDER_PLATFORM)
```
Gnimuc, line 131:
```suggestion
    Artifacts.ensure_artifact_installed(get_gcc_shard_key(triple, version), ARTIFACT_TOML_PATH[])
```

This is fine because the dummy argument `platform` is ignored(not used for selecting/filtering artifacts like it is used in `artifact_meta`, etc.) in the current impl of `ensure_artifact_installed`.

src/shards/JLLEnvs.jl
```diff
@@ -8,12 +8,18 @@ include("utils.jl")
 
 const JLL_ENV_SHARDS = Dict{String,Any}()
 
+const ARTIFACT_TOML_PATH = Ref{String}()
+
+# The platform used in Artifacts.toml
+const BINARY_BUILDER_PLATFORM = Base.BinaryPlatforms.Platform("x86_64", "linux"; libc="musl")
+
```
Gnimuc, line 15:
```suggestion
```

IIRC, there is an "Any" platform that can be used if we do need to specify `platform` in the future.

Gnimuc (changes requested):

src/shards/JLLEnvs.jl
melonedo: No, subtypes of `AbstractPlatform` are only `Platfrom`, `UnknownPlatform` and the ones defined in Pkg, and `UnknownPlatform` does not match any platforms.

src/shards/JLLEnvs.jl
melonedo: I also hoped it would have works, but it defaults to `HostPlatform()`😢.
https://github.com/JuliaLang/Pkg.jl/blob/e31a3dc77201e1c7c469f6d4572c521f93fefb20/src/Artifacts.jl#L397-L402

Gnimuc: What I meant was to directly call this method `ensure_artifact_installed(name::String, meta::Dict, artifacts_toml::String;...)`

Issue #376, opened 2022-02-16
problem with `resolve_dependency!(dag::ExprDAG, node::ExprNode{<:AbstractMacroNodeType, options)`


jkozdon:
I've been trying to revive some `PETSc` wrappers I started a long time ago, and upon updating `Clang.jl` to the latest release I ran into a problem caused by e6022d2f710563bfb98cd9738932f0c7c01ad885.

When this commit is applied, the [`RemoveCircularReference`](https://github.com/JuliaInterop/Clang.jl/blob/8ff4b1bc9603fa6a80b7b1c12f8bd2bf6785e1ce/src/generator/passes.jl#L341-L404) pass just loops forever.

If I revert these changes, even with the latest master branch, the code will at least complete for a simple test case.

I've tried to strip things down pretty minimally in this [gist](https://gist.github.com/jkozdon/7b25895dce3e4f2fb94f2c361128c5c7) for what I am running.

Gnimuc:
cc @melonedo

melonedo:
This is because `PETSc_jll/include/pestclog.h` defines macros that "hide" the implementation, which looks like circular reference.
```c
#define foo(a,b,c) log(a,b,c); foo(a,b,c)
```

PR #378, opened 2022-02-17
Fix issue #376

melonedo:
Fix issue #376


Gnimuc:
Would be great to add https://github.com/JuliaInterop/Clang.jl/issues/376#issuecomment-1042629419 as a test case.

PR #384, opened 2022-08-24
Fix #233

### LINKED ISSUE #233: Unnamed struct members are skipped

Gnimuc:
cc @cdsousa


cdsousa:
Strange, if I checkout that branch and run the tests, that new test will also pass for me.

However, running this simple script in REPL fails:

```julia

using Clang
using Clang.Generators

header_content = """
struct Meow {
    int blah;
    union {
        float either1;
        int either2;
    };
    float moo;
};
"""

header_path = tempname()*".h"
write(header_path, header_content)

ctx = create_context(header_path, get_default_args())
build!(ctx)
```

cdsousa:
Wow! If I do  `header_path = "unnamed.h"` then it works 😲
Actually, it works if, and only if, the filename starts with "unnamed"... What the heck!

cdsousa:
Alright, it seems it works only if the "unnamed" is present in the filename of in the names of the parent types.
Probably that's related to https://github.com/JuliaInterop/Clang.jl/blob/master/src/generator/Generators.jl#L39
Is that by design?

cdsousa:
Note that (as I edited in https://github.com/JuliaInterop/Clang.jl/pull/384#issuecomment-1225915158) it suffices for the filename to include the string "unammed" somewhere (or that be present on any name of the type hierarchy).
So it is expected that "test_unnamed.h" also passes the test.

Gnimuc:
> Alright, it seems it works only if the "unnamed" is present in the filename or in the names of the parent types. Probably that's related to https://github.com/JuliaInterop/Clang.jl/blob/master/src/generator/Generators.jl#L39 Is that by design?

This is just a workaround for detecting anonymous record types. This is a limitation of libclang. I guess the ultimate solution is to rework the generator on top of ClangCompiler.jl.

cdsousa:
Thank you! Thanks for all your work!

Issue #380, opened 2022-05-27
Respect `output_ignorelist` for macro definitions


sloede:
When trying to generate bindings for the [p4est](https://github.com/cburstedde/p4est) headers, adding `sc_MPI` to the `output_ignorelist` setting in the `generator.toml` has no effect: Macros of the type
```c
#define sc_MPI_COMM_WORLD          MPI_COMM_WORLD
```
are still translated to
```julia
const sc_MPI_COMM_WORLD = MPI_COMM_WORLD
```
It seems like macro printing ignores the `output_ignorelist`, or am I missing something? If I am right, can this be easily fixed?

Gnimuc:
I can not reproduce this. 

It should be handled here: https://github.com/JuliaInterop/Clang.jl/blob/9931c38d6d699f49923206a2c1bece819ef017c7/src/generator/passes.jl#L957

```
julia> using Clang.Generators

julia> args = get_default_args()
6-element Vector{String}:
 "-isystem/Users/gnimuc/.julia/ar" ⋯ 61 bytes ⋯ "64-apple-darwin14/4.8.5/include"
 "-isystem/Users/gnimuc/.julia/ar" ⋯ 67 bytes ⋯ "le-darwin14/4.8.5/include-fixed"
 "-isystem/Users/gnimuc/.julia/ar" ⋯ 47 bytes ⋯ "9/x86_64-apple-darwin14/include"
 "-isystem/Users/gnimuc/.julia/ar" ⋯ 60 bytes ⋯ "e-darwin14/sys-root/usr/include"
 "-isystem/Users/gnimuc/.julia/ar" ⋯ 74 bytes ⋯ "-root/System/Library/Frameworks"
 "--target=x86_64-apple-darwin14"

shell> cat test.h
#define sc_MPI_COMM_WORLD          MPI_COMM_WORLD
typedef int cs_long_t;
typedef void * cs_dl;

cs_long_t cs_dl_fkeep (cs_dl *A, cs_long_t (*fkeep) (cs_long_t, cs_long_t, double, void *), void *other) ;

julia> ctx = create_context("./test.h", args, Dict("general"=>Dict{String,Any}("output_ignorelist"=>["sc_MPI_COMM_WORLD"])))
[ Info: Parsing headers...
Context(...)

julia> build!(ctx)
[ Info: Processing header: test.h
[ Info: Building the DAG...
[...]

PR #385, opened 2022-09-28
add support for ignore by regex

t-bltg:
I'm trying to generate `MPI` signatures in https://github.com/JuliaParallel/MPI.jl/pull/644.

It would be convenient to filter output by regex, since `output_ignorelist` requires listing everything by hand (and this would take more time than writing this PR).

E.g. this works in my `generator.toml`:
```
output_ignoreregex = ["^PMPI_", "^QMPI_"]
```

### DIFF
--- a/src/generator/passes.jl
+++ b/src/generator/passes.jl
@@ -852,11 +852,11 @@ end
 
 function should_exclude_node(node, ignoreregex, ignorelist, exclusivelist)
     str_node = string(node.id)
-    for re in ignoreregex
+    for re ∈ ignoreregex
         match(re, str_node) === nothing || return true
     end
     str_node ∈ ignorelist && return true
-    if exclusivelist !== nothing && !(str_node in exclusivelist)
+    if exclusivelist !== nothing && str_node ∉ exclusivelist
         return true
     end
     return false
--- a/src/generator/passes.jl
+++ b/src/generator/passes.jl
@@ -850,8 +850,11 @@ function (x::Audit)(dag::ExprDAG, options::Dict)
     return dag
 end
 
-function should_exclude_node(node, ignorelist, exclusivelist)
+function should_exclude_node(node, ignoreregex, ignorelist, exclusivelist)
     str_node = string(node.id)
+    for re in ignoreregex
+        match(re, str_node) === nothing || return true
+    end
     str_node ∈ ignorelist && return true
     if exclusivelist !== nothing && !(str_node in exclusivelist)
         return true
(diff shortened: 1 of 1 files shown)


Gnimuc:
It would be great if we could add regex support to `output_ignorelist` instead of introducing a new config entry.

Gnimuc:
X-ref: https://github.com/JuliaInterop/Clang.jl/issues/380

t-bltg:
> It would be great if we could add regex support to output_ignorelist

The problem is that the `.toml` cannot parse regex string macros:

```julia
ERROR: TOML Parser error:
none:5:21 error: unexpected start of value
  output_ignorelist = [r"^PMPI_", r"^QMPI_", r"^MPIX_", r"^PMPIX_", r"^QMPIX_", r"^MPI_T_", r"^MPI_Session_"]
                      ^                                                                                       
Stacktrace:
 [1] parse
   @ ./toml_parser.jl:441 [inlined]
```

Thus, we cannot distinguish regexes from regular string in `output_ignorelist`, thus breaking the current behavior.

Please advise on the implementation ?

Gnimuc:
Is it feasible to treat everything as regex string?

Gnimuc:
I can not come up with an ambitious case.

t-bltg:
> Is it feasible to treat everything as regex string?

I guess I can try. Would that imply `exclusivelist` should become one too ?

src/generator/passes.jl
```diff
@@ -879,7 +881,7 @@ function (x::FunctionPrinter)(dag::ExprDAG, options::Dict)
     general_options = get(options, "general", Dict())
     log_options = get(general_options, "log", Dict())
     show_info = get(log_options, "FunctionPrinter_log", x.show_info)
-    ignorelist = get(general_options, "output_ignorelist", get(general_options, "printer_blacklist", []))
+    ignorelist = Regex.(get(general_options, "output_ignorelist", get(general_options, "printer_blacklist", [])))
```
Gnimuc, line 884: Better to do the type conversion in `should_exclude_node`, like:

```
for item ∈ ignorelist
      match(Regex(item), str_node) === nothing || return true
end
```

t-bltg: Done.

Gnimuc:
Thanks!

t-bltg:
Amazing.

Would it be possible to tag a new version so that we can add a valid `[compat]` entry for `Clang.jl` in https://github.com/JuliaParallel/MPI.jl/pull/644 ?

PR #386, opened 2022-09-30
allow documentation `callback`s

t-bltg:
Provide a way to provide a documentation callback, when parsed c documentation is absent, implementation is `tbd`:

```julia

options = load_options(joinpath(@__DIR__, "generator.toml"))

options["general"]["callback_documentation"] = node -> String["This is the docstring generated for $(node.id)"]
```

For https://github.com/JuliaParallel/MPI.jl/pull/652.


src/generator/documentation.jl
```diff
@@ -45,8 +44,8 @@ function print_documentation(io::IO, cursor::CLCursor, indent, options, members:
     end
 end
 
-function get_prototype(node)
```
t-bltg, line 48: Was this functional before ? (i.e. a cursor was given, not a node...)

src/generator/documentation.jl
```diff
@@ -14,15 +9,19 @@ function print_documentation(io::IO, cursor::CLCursor, indent, options, members:
...
-        return
+        if (callback_documentation = get(options, "callback_documentation", nothing)) ≢ nothing
+            doc = callback_documentation(node)
+        else
+            return
+        end
     end
     
-    if show_c_function_prototype && cursor isa Clang.CLFunctionDecl
-        prototype = ["### Prototype", "```c", get_prototype(cursor), "```"]
```
t-bltg, line 25: was this functional before ? (i.e. `get_prototype` signature states `node`, not `cursor`)

Gnimuc: cc @melonedo

src/generator/documentation.jl
```diff
@@ -14,15 +9,19 @@ function print_documentation(io::IO, cursor::CLCursor, indent, options, members:
...
     if extract_c_comment_style == "doxygen"
-        doc = format_doxygen(cursor, options, members)
+        doc = format_doxygen(node.cursor, options, members)
     elseif extract_c_comment_style == "raw"
-        doc = format_raw(cursor, members)
+        doc = format_raw(node.cursor, members)
     else
-        return
+        if (callback_documentation = get(options, "callback_documentation", nothing)) ≢ nothing
+            doc = callback_documentation(node)
```
Gnimuc, line 17: This config entry can not be put in the toml config file, right?

t-bltg: No, it has to be a callback.

t-bltg:
@gnimuc, I've added specific tests for the configuration we use in `MPI.jl`.

Those also cover the previous PR: https://github.com/JuliaInterop/Clang.jl/pull/385.

test/runtests.jl
```diff
@@ -1,5 +1,9 @@
 using Clang
 using Test
+using Pkg
 
 include("generators.jl")
+
+include("test_mpi.jl")
+Pkg.build("CMake")
```
t-bltg, line 8: Local tests were giving `CMake` errors.

Gnimuc: Why use CMake here?

t-bltg: No, it's for the following test `test_bitfield.jl` which uses `CMake`.

melonedo: IIRC, building CMake.jl should be done when installing a package, not when testing it.

t-bltg: Yeah well, this errors out locally. Will try to remove for remote `ci`.

test/mpi/mpi.h
```diff
@@ -0,0 +1,3 @@
+typedef int MPI_Comm;
+int MPI_Barrier(MPI_Comm comm);
+int PMPI_Barrier(MPI_Comm comm);
```
Gnimuc, line 3: There is no need to build any binary to test this header.

t-bltg: What do you mean ?

Gnimuc: I mean you can directly use this file path for `headers` instead of moving this header to a build directory.

test/runtests.jl
t-bltg: I've removed it, CI looks ok.

test/test_mpi.jl
```diff
@@ -0,0 +1,21 @@
+using Clang.Generators
+
+@testset "MPI.jl" begin
+    @info "Testing generator for JuliaParallel/MPI.jl"
+    mkpath("build_mpi")
```
Gnimuc, line 5: Following my comments above, I think there is no need to use this.

test/mpi/mpi.h
t-bltg: This was for `.gitignore`ing the auto-generated files. We don't need them in the repo.

test/test_mpi.jl
Gnimuc: This name is very confusing. You could use another directory name like `gen_mpi`.

test/mpi/mpi.h
t-bltg: Solved by writing to the test dir instead.

test/test_mpi.jl
t-bltg: Solved.

src/generator/documentation.jl
melonedo: Not sure whether it works now, but if it confuses you, `node.cursor.cursor` and `node.cursor` both work for some functions. This function used to accept ExprNode so it was possible I did not adjust it.

t-bltg: You tell me, should I restore to the previous (being cautious) i.e. `get_prototype(node.cursor)` instead ?

Gnimuc: I think we can fix it in this way:

```
get_prototype(node) = get_prototype(node.cursor)

function get_prototype(cursor::CLCursor)
    code = Clang.getSourceCode(cursor)
    code = replace(code, r"/\*.*?\*/|//.*?$"m=>"")
    code = replace(code, r"\s+"m=>" ")
    "$code;"
end
```

t-bltg: Thanks, pushed in 3d792337c49678506c1f9ea559c039a2b697ee81.

melonedo: get_prototype is only called here (actually, it was factored out to shorten the function body), so you are free to make any change as long as it works.

Gnimuc:
I will do a squash merge if there are no additional changes.

t-bltg:
I'm done, pending CI does not fail.

t-bltg:
Thanks all for the valuable comments ;)

PR #390, opened 2022-10-25
Fix #389

Gnimuc:


melonedo (commented):
Isn't Base.length defined for TokenList? There seems to be no need to split them.

Gnimuc:
Fixed. Thanks!

Gnimuc:
cc @maleadt

PR #398, opened 2022-11-20
Unix C library compatibility

notinaboat:
Changes to enable wrapping of the system C library on mac OS and Linux.

See commit logs for detail.

Currently I am wrapping: `<errno.h>`, `<string.h>`, `<limits.h>`, `<stdlib.h>`, `<pthread.h>`, `<sys/ioctl.h>`, `<termios.h>`, `<fcntl.h>`, `<poll.h>`, `<sys/epoll.h>` `<unistd.h>`, `<sys/stat.h>`, `<net/if.h>`, `<sys/socket.h>`, `<signal.h>`, `<sys/wait.h>`, `<sys/syscall.h>` and `<spawn.h>`.

... and i'm using these options:
```julia
   Dict{String,Any}(
            "general" => Dict{String,Any}(
                "is_local_header_only" => false,
                "auto_mutability" => true,
                "auto_mutability_includelist" => ["termios"],
                "library_name" => ""
            ),
            "codegen" => Dict{String,Any}(
                "use_ccall_macro" => true,
                "wrap_variadic_function" => false
            ),
            "codegen.macro" => Dict{String,Any}(
                "macro_mode" => "aggressive",
                "ignore_pure_definition" => true
            )
        )
```

See https://github.com/notinaboat/UnixIO.jl/blob/c03f1695999666ab95177b5e1713e2fe1cd1eae0/packages/UnixIOHeaders/src/UnixIOHeaders.jl#L162-L188

### DIFF
--- a/src/generator/audit.jl
+++ b/src/generator/audit.jl
@@ -73,7 +73,7 @@ function sanity_check(dag::ExprDAG, options::Dict)
         elseif is_function(idn)
             ifile, iline, icol = get_file_line_column(idn.cursor)
             tfile, tline, tcol = get_file_line_column(tn.cursor)
-            error("sanity check failed. [REASON]: use the same name $tk for struct $(tn.cursor) at $tfile:$tline:$tcol and function $(idn.cursor) at $ifile:$iline:$icol. The C code use the same name for structs and functions, which is not a good code-style, please fix it in the upstream.")
+            @warn "sanity check failed. [REASON]: use the same name $tk for struct $(tn.cursor) at $tfile:$tline:$tcol and function $(idn.cursor) at $ifile:$iline:$icol. The C code use the same name for structs and functions, which is not a good code-style, please fix it in the upstream."
         else
             error("sanity check failed. please file an issue to Clang.jl.")
         end
--- a/src/generator/passes.jl
+++ b/src/generator/passes.jl
@@ -32,7 +32,15 @@ function (x::CollectTopLevelNode)(dag::ExprDAG, options::Dict)
         header_name = normpath(spelling(tu_cursor))
         @info "Processing header: $header_name"
         for cursor in children(tu_cursor)
-            file_name = normpath(get_filename(cursor))
+            file_name = get_filename(cursor)
+
+            # Ignore items where no source file is available.
+            # e.g. built-in defines like `__llvm__` and `__clang__`.
+            if !is_local_only && isempty(file_name)
+                continue
+            end
+
+            file_name = normpath(file_name)
             if is_local_only && header_name != file_name && file_name ∉ x.dependent_headers
                 if any(sysdir -> startswith(file_name, sysdir), x.system_dirs)
                     collect_top_level_nodes!(dag.sys, cursor, general_options)
--- a/src/generator/codegen.jl
+++ b/src/generator/codegen.jl
@@ -87,9 +87,10 @@ function emit!(dag::ExprDAG, node::ExprNode{FunctionProto}, options::Dict; args.
     use_ccall_macro = get(options, "use_ccall_macro", false)
     if use_ccall_macro
         name_type_pairs = [Expr(:(::), n, t) for (n, t) in zip(arg_names, args)]
+        library_func = library_expr == nothing ? func_name : :($library_expr.$func_name)
         body = Expr(:macrocall, Symbol("@ccall"), nothing,
                     Expr(:(::), Expr(:call,
-                                        :($library_expr.$func_name),
+                                        library_func,
                                         name_type_pairs...,
                                     ),
                          ret_type


src/generator/audit.jl
```diff
@@ -73,7 +73,7 @@ function sanity_check(dag::ExprDAG, options::Dict)
         elseif is_function(idn)
             ifile, iline, icol = get_file_line_column(idn.cursor)
             tfile, tline, tcol = get_file_line_column(tn.cursor)
-            error("sanity check failed. [REASON]: use the same name $tk for struct $(tn.cursor) at $tfile:$tline:$tcol and function $(idn.cursor) at $ifile:$iline:$icol. The C code use the same name for structs and functions, which is not a good code-style, please fix it in the upstream.")
+            @warn "sanity check failed. [REASON]: use the same name $tk for struct $(tn.cursor) at $tfile:$tline:$tcol and function $(idn.cursor) at $ifile:$iline:$icol. The C code use the same name for structs and functions, which is not a good code-style, please fix it in the upstream."
```
Gnimuc, line 76: I'm wondering what the generated code looks like. Is it possible to map these cases on the Julia side? Julia does not support duplicated struct/function identifiers.

src/generator/resolve_deps.jl
```diff
@@ -116,6 +116,8 @@ function resolve_dependency!(dag::ExprDAG, node::ExprNode{TypedefDefault}, optio
         push!(node.adj, dag.ids[leaf_ty.sym])
     elseif haskey(dag.ids_extra, leaf_ty.sym)
         # pass
+    elseif startswith(spelling(getTypeDeclaration(ty)), "__builtin_")
```
Gnimuc, line 119: Is it possible to solve this problem by using `@add_def` macro?

src/generator/passes.jl
```diff
@@ -32,7 +32,15 @@ function (x::CollectTopLevelNode)(dag::ExprDAG, options::Dict)
         header_name = normpath(spelling(tu_cursor))
         @info "Processing header: $header_name"
         for cursor in children(tu_cursor)
-            file_name = normpath(get_filename(cursor))
+            file_name = get_filename(cursor)
+
+            # Ignore items where no source file is available.
+            # e.g. built-in defines like `__llvn__` and `__clang__`.
```
Gnimuc, line 38: I guess these symbols should be defined in the header files of the compiler's include directory.

Gnimuc:
Is there any reason why you choose not to use [Ygg system headers](https://github.com/JuliaInterop/Clang.jl/blob/master/src/shards/JLLEnvs.jl)?

src/generator/audit.jl
notinaboat: The code looks like:

```julia
struct stat
    st_dev::dev_t
    st_mode::mode_t
    st_nlink::nlink_t
    st_ino::__darwin_ino64_t
    st_uid::uid_t
    st_gid::gid_t
    st_rdev::dev_t
    st_atimespec::timespec
    st_mtimespec::timespec
    st_ctimespec::timespec
    st_birthtimespec::timespec
    st_size::off_t
    st_blocks::blkcnt_t
    st_blksize::blksize_t
    st_flags::__uint32_t
    st_gen::__uint32_t
    st_lspare::__int32_t
    st_qspare::NTuple{2, __int64_t}
end
```

```julia
function stat(arg1, arg2)
    @ccall stat(arg1::Ptr{Cchar}, arg2::Ptr{stat})::Cint
end
```

It seems that Julia does not have any problem with struct/function aliasing, e.g.:

```julia
% julia18
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.2 (2022-09-29)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> struct foo
       a::Int
       end

julia> foo(a,b,c) = a+b+c
foo

julia> foo(4)
foo(4)

julia> typeof(foo(4))
foo

julia> foo(1,2,3)
6
```

Gnimuc: That is actually an outer constructor of `foo`.

Gnimuc:
```suggestion
            @warn "sanity check failed. [REASON]: use the same name $tk for struct $(tn.cursor) at $tfile:$tline:$tcol and function $(idn.cursor) at $ifile:$iline:$icol. The C code use the same name for structs and functions, which is not a good code-style, the behavior of the generated is undefined."
```

src/generator/resolve_deps.jl
notinaboat: I originally tried `"output_ignorelist" => ["__darwin_va_list"]`, but that only suppresses printing.

I wasn't aware of `@add_def`, but it doesn't do quite what I want either. It creates a new type in the calling module. e.g. ...

```julia
julia> module Foo
       using Clang
       using Clang.Generators
       @add_def bar
       end
Main.Foo

julia> Foo.bar isa DataType
true
```

What I really want is to ignore `__darwin_va_list` completely. 

It looks like I can solve the problem by doing this:
```julia
    Generators.add_definition(:__builtin_va_list => Generators.JuliaUnsupported())
```

I will remove this commit from the PR.

src/generator/passes.jl
notinaboat: No, these symbols don't come from header files, they are hard-coded in the source code of the compiler line this:
```c++
  // Compiler version introspection macros.
  Builder.defineMacro("__llvm__");  // LLVM Backend
  Builder.defineMacro("__clang__"); // Clang Frontend
```
See  https://github.com/llvm/llvm-project/blob/8699efba6dc8d905d144727eb10ff6782376bb87/clang/lib/Frontend/InitPreprocessor.cpp#L744-L746

notinaboat:
> Is there any reason why you choose not to use [Ygg system headers](https://github.com/JuliaInterop/Clang.jl/blob/master/src/shards/JLLEnvs.jl)?

Those headers appear to be out of date.

For example, I use the POSIX spawn interface here: https://github.com/notinaboat/UnixIO.jl/blob/julia18/src/UnixIO.jl#L1282-L1284

But, the `POSIX_SPAWN_SETSID` symbol is not defined in the Ygg headers...

```julia
julia> using Clang

julia> dirs =Clang.JLLEnvs.get_system_dirs("x86_64-apple-darwin14")
5-element Vector{String}:
 "/Users/samoconnor/.julia/artifa" ⋯ 57 bytes ⋯ "64-apple-darwin14/4.8.5/include"
 "/Users/samoconnor/.julia/artifa" ⋯ 63 bytes ⋯ "le-darwin14/4.8.5/include-fixed"
 "/Users/samoconnor/.julia/artifa" ⋯ 43 bytes ⋯ "9/x86_64-apple-darwin14/include"
 "/Users/samoconnor/.julia/artifa" ⋯ 56 bytes ⋯ "e-darwin14/sys-root/usr/include"
 "/Users/samoconnor/.julia/artifa" ⋯ 70 bytes ⋯ "-root/System/Library/Frameworks"

julia> dirs[1]
"/Users/samoconnor/.julia/artifacts/303e65f6896f295459c81aa311b5eb993f702d89/lib/gcc/x86_64-apple-darwin14/4.8.5/include"
```

e.g. the file `spawn.h` is found...
```bash
% find /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/ -name spawn.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk//usr/include/spawn.h
...
```

But the symbol `POSIX_SPAWN_SETSID` is not found...
```bash
% find /Users/samoconnor/.julia/artifacts/303e65f6896f295459c81aa311b5eb993f702d89 -name spawn.h | xargs grep POSIX_SPAWN_SETSID
% 
```

Compare to searching under `Applications/Xcode.app` ...
```bash
% find /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/ -name spawn.h | xargs grep POSIX_SPAWN_SETSID
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk//usr/include/sys/spawn.h:#define POSIX_SPAWN_SETSID              0x0400
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk//System/Library/Frameworks/Kernel.framework/Versions/A/Headers/sys/spawn.h:#define POSIX_SPAWN_SETSID              0x0400
```

notinaboat:
... also, I'm a little uneasy about using platform headers from `https://github.com/phracker/MacOSX-SDKs` I would rather use official headers form the platform vendor.

notinaboat:
Background FYI:

My motivation for creating UnixIO.jl came from working on an embedded system (R'Pi-linux) and having trouble with unreliable serial port IO.
I could write a tiny C program to open `/dev/ttyS0` and read/write with 100% reliability, but from Julia it was unreliable.
I started trying to find the root cause, but that was complicated by all the intermediate layers in `Base.IO` and [`libuv`](https://github.com/libuv/libuv).
It seems that the libuv project simply does really intend to support serial ports: https://github.com/libuv/libuv/issues/1970

So, the intention of UnixIO.jl is:
 - I just want to talk directly to the native OS C library on my machine.
 - I want to avoid any risk that might be introduced by 3rd party intermediate layers.

Gnimuc:
> Those headers appear to be out of date.
> 
> For example, I use the POSIX spawn interface here: https://github.com/notinaboat/UnixIO.jl/blob/julia18/src/UnixIO.jl#L1282-L1284
> 
> But, the `POSIX_SPAWN_SETSID` symbol is not defined in the Ygg headers...

The default gcc version used in Clang.jl is v0.4.8.5, which can be updated to a newer one. But in your use case, I think it's better to just use headers on your local system.

> Background FYI:
> 
> My motivation for creating UnixIO.jl came from working on an embedded system (R'Pi-linux) and having trouble with unreliable serial port IO. I could write a tiny C program to open `/dev/ttyS0` and read/write with 100% reliability, but from Julia it was unreliable. I started trying to find the root cause, but that was complicated by all the intermediate layers in `Base.IO` and [`libuv`](https://github.com/libuv/libuv). It seems that the libuv project simply does really intend to support serial ports: [libuv/libuv#1970](https://github.com/libuv/libuv/issues/1970)
> 
> So, the intention of UnixIO.jl is:
> 
> * I just want to talk directly to the native OS C library on my machine.
> * I want to avoid any risk that might be introduced by 3rd party intermediate layers.

Make sense to me now. Thanks for the PR. :)

PR #408, opened 2022-12-16
fix `typedef <function prototype>`

t-bltg:
https://github.com/JuliaParallel/MPI.jl/pull/693?notification_referrer_id=NT_kwDOAMzS8LM1MTA3NTQ4MzA0OjEzNDIzMzQ0#issuecomment-1354933169


t-bltg:
@Gnimuc, this seems to fix https://github.com/JuliaParallel/MPI.jl/issues/688.

The new generated file in https://github.com/JuliaParallel/MPI.jl/pull/693 seem to correctly use `Ptr{<some typedefed function>}` instead of `Ptr{Cvoid}`.

Gnimuc:
As #403 introduced a breaking change, I'm going to tag the next version as `0.17.0`.

t-bltg:
Thanks 🙏

PR #419, opened 2023-03-09
Try to fix `getNumArguments` for `TypeRef`-ed `FunctionTemplateDecl`s

Gnimuc:
I don't know how to craft an MWE to test this PR, but this may be the solution.

cc @giordano


giordano:
I have the feeling this might be a problem in libclang, because `Clang.LibClang.clang_Cursor_getNumArguments(method)` gives me `-1` on the method where I expect >1 arguments

Gnimuc:
it looks like `FunctionTemplateDecl` can not be `dyn_cast` to a typical `FunctionDecl` in the newer Clang. I'm closing this PR as it's not the correct way to fix the problem.

PR #421, opened 2023-03-11
`detect_headers`: sort headers list

t-bltg:
Fix https://github.com/Gnimuc/ClangCompiler.jl/issues/24.


Gnimuc:
If Julia's `Set` is not deterministic between versions, then it's not a problem of Clang. I'd like to keep the dependency graph minimal.

t-bltg:
No, it's non-deterministic for a single julia version:
```
$ julia generator.jl 1.9
$ julia generator.jl 1.9  # re-run, signatures changed
```

Gnimuc:
Is it same with 1.8?

src/generator/resolve_deps.jl
```diff
@@ -182,7 +182,7 @@ resolve_dependency!(dag::ExprDAG, node::ExprNode{<:AbstractUnionNodeType}, optio
 # at least sort macros according to identifiers, otherwise the order is terribly wrong
 function resolve_dependency!(dag::ExprDAG, node::ExprNode{<:AbstractMacroNodeType}, options)
     # macro name and its arguments are not dependencies
-    args = Set{Symbol}((node.id,))
+    args = OrderedSet{Symbol}((node.id,))
```
Gnimuc, line 185: this container is used as a hash table, the order doesn't matter at all.

Gnimuc:
> How about using a Vector and use unique! afterwards ?
> Or use sort ?

yes, this is a good idea.

src/generator/resolve_deps.jl
t-bltg: Oh, yeah noticed while reworking.

PR #424, opened 2023-03-17
Cleanup CXFile support and add unique_id

habemus-papadum:
This is somewhat invasive, but hopefully not too scary. 

It will help for adding support for modules which are uniqued by their ast file. I included the commit from #423 which is pretty simple.  I can close that PR if you  take this onn or I can rebase this one later.

### DIFF
--- a/src/Clang.jl
+++ b/src/Clang.jl
@@ -33,6 +33,9 @@ end
 
 include("string.jl")
 
+include("file.jl")
+export CLFile, name, unique_id, get_filename
+
 include("index.jl")
 export Index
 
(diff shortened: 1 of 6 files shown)


Gnimuc:
As this PR introduces breaking changes, we should add a guidance somewhere to let users know how to upgrade. 

Another way is to rename `get_file_line_column` to `get_filecursor_line_column` and deprecate `get_file_line_column` in v0.18.

habemus-papadum:
Thinking it through, I think it is cleaner to wrap CXFile in a CLFile, and to not make breaking changes.  I will update the PR

Gnimuc:
This should be added in v0.18.

PR #428, opened 2023-03-30
Use visitor for searching `CLFunctionTemplate`'s `CLParmDecl`

Gnimuc:


giordano:
This seems to be working well for me! :rocket: This is awesome, thanks a lot!  Is there any chance to have a test for this, to avoid regressions in the future?

PR #456, opened 2023-11-25
Add an option for skipping static functions

### LINKED ISSUE #455: Does Clang.jl have an option to skip `static` functions?
Related to #267, but recently resurfaced in https://github.com/trixi-framework/P4est.jl/issues/99. We know that it is not possible to make `static` functions callable from Julia (at least not trivially; since that would either require source-to-source translation or modifying the original library). However, it would be great if we had a mechanism to prevent them from being included in the generated Julia API.

cc @ranocha @Blumenkranz

Gnimuc:
close #455 

cc @sloede

### DIFF
--- a/gen/generator.toml
+++ b/gen/generator.toml
@@ -169,6 +169,9 @@ field_access_method_list = []
 # prevent the generated symbols from conflicting with the symbols defined and exported in Base.
 function_argument_conflict_symbols = []
 
+# if set to true, static functions will be ignored
+skip_static_functions = false
+
 # emit constructors for all custom-layout structs like bitfield in the list,
 # or set to `true` to do so for all such structs
 add_record_constructors = []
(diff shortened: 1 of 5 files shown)


JoshuaLampert:
Thanks for this quick addition @Gnimuc! However, for me, static functions are only skipped if `skip_static_functions = true` in the generator.toml is **before** the `[codegen]` block other than suggested in the [example file](https://github.com/JuliaInterop/Clang.jl/blob/master/gen/generator.toml#L173). Is that to be expected and if yes, could this be adjusted in the example generator.toml?

Gnimuc:
> Thanks for this quick addition @Gnimuc! However, for me, static functions are only skipped if `skip_static_functions = true` in the generator.toml is **before** the `[codegen]` block other than suggested in the [example file](https://github.com/JuliaInterop/Clang.jl/blob/master/gen/generator.toml#L173). Is that to be expected and if yes, could this be adjusted in the example generator.toml?

Oops! That's a miss.

JoshuaLampert:
Alright, thanks for the quick reply and the quick fix!

PR #461, opened 2024-01-02
Move the tests to use ReTest.jl

JamesWrigley:
This allows for a faster workflow when developing the package. Also made a few paths in the tests explicitly relative to the `test/` directory instead of the PWD (since that may not be `test/` when running the tests interactively).

(as discussed on Slack)


JamesWrigley:
Waiting on https://github.com/JuliaTesting/ReTest.jl/pull/55.

Gnimuc:
Do we need this to be a part of v0.18?

JamesWrigley:
Nah, it's very low-priority.

JamesWrigley:
Ok, this should be ready to go now :) I also saw errors in CI from using the latest BB, so I set compat bounds in 3401cd9.

Issue #464, opened 2024-01-13
Unable to install Clang.jl on Julia master for 1.11-dev - needs new libLLVM_jll


ViralBShah:
```
(@v1.11) pkg> add Clang@0.17.8
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package libLLVM_jll [8f36deef]:
 libLLVM_jll [8f36deef] log:
 ├─possible versions are: 16.0.6 or uninstalled
 └─found to have no compatible versions left with Clang_jll [0ee61d77]
   └─Clang_jll [0ee61d77] log:
     ├─possible versions are: 9.0.1 - 17.0.6 or uninstalled
     └─restricted by compatibility requirements with Clang [40e3b903] to versions: 12.0.0 - 15.0.7
       └─Clang [40e3b903] log:
         ├─possible versions are: 0.7.0 - 0.17.8 or uninstalled
         └─restricted to versions 0.17.8 by an explicit requirement, leaving only versions: 0.17.8```

**Comment by ViralBShah** [comment]
Perhaps the same as #234?

**Comment by ViralBShah** [comment]
I'm using a very old Clang.jl, it turns out.

**Comment by ViralBShah** [comment]
The issue seems to be the lack of availability of an updated `libLLVM_jll`, which is forcing Clang.jl down to an old version.

```
(@v1.11) pkg> add Clang@0.17.8
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package libLLVM_jll [8f36deef]:
 libLLVM_jll [8f36deef] log:
 ├─possible versions are: 16.0.6 or uninstalled
 └─found to have no compatible versions left with Clang_jll [0ee61d77]
   └─Clang_jll [0ee61d77] log:
     ├─possible versions are: 9.0.1 - 17.0.6 or uninstalled
     └─restricted by compatibility requirements with Clang [40e3b903] to versions: 12.0.0 - 15.0.7
       └─Clang [40e3b903] log:
         ├─possible versions are: 0.7.0 - 0.17.8 or uninstalled
         └─restricted to versions 0.17.8 by an explicit requirement, leaving only versions: 0.17.8
```

PR #466, opened 2024-01-21
Attempt to follow the Julia docstring conventions

JamesWrigley:
In particular, the docs recommends `# Arguments` instead of `### Parameters`: https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation

They also use a single '#' for other section headings, though in practice Documenter.jl seems to render them exactly the same anyway.


JamesWrigley:
Ah, I see nightly tests are broken because of #464 :thinking:

PR #473, opened 2024-03-06
Documentation test fixes

JamesWrigley:
Supersedes #472. On that PR the windows x64 CI was hanging :thinking: Let's see if it was just a fluke or if it happens again, if it does then maybe something is detecting the `REPL` module and creating a prompt.


JamesWrigley:
Ok... we are finally back in the green :partying_face: The problem was actually a bug introduced by me in #458, the tests were calling the cmake binary directly instead of using the `cmake()` cmd which also sets appropriate environment variables. Fixed that in 562d430.

Gnimuc:
Thanks! 🎉

PR #477, opened 2024-03-08
Add support for checking `library_names` in the generator auditor

### LINKED ISSUE #476: warning for `library_name`
Generating the SuiteSparse wrappers, I got:

```
└ @ Clang.Generators ~/.julia/packages/Clang/kNxXb/src/generator/audit.jl:16
```

However, I do have `library_names`, and maybe the audit can check for that too.

```
generator.toml:library_names = {"SuiteSparse_config.h" = "libsuitesparseconfig", "SuiteSparseQR_C.h" = "libspqr", "umfpack.h" = "libumfpack", "cholmod.h" = "libcholmod", "amd.h" = "libamd"}
```

JamesWrigley:
This fixes spurious warnings when `library_names` is set instead of `library_name`.

Fixes #476.

### DIFF
--- a/src/generator/audit.jl
+++ b/src/generator/audit.jl
@@ -13,7 +13,7 @@ function audit_library_name(dag::ExprDAG, options::Dict)
         end
     else
         general_options["library_name"] = "libxxx"
-        @warn "default libname: \":libxxx\" is being used, did you forget to set `library_name` in the toml file?"
+        @warn "default libname: \":libxxx\" is being used, did you forget to set `library_name` in the toml file? It's safe to ignore this warning if you are using `library_names` as an exhaustive list (if so, set `library_name` to a dummy value to supress it)."
     end
     if haskey(options, "general")
         merge!(options["general"], general_options)


Gnimuc:
`library_name` is the fallback libname for those headers that are not captured by `library_names`.

```
# library_name = ">/+-;'[]{}()!@#$%^&*~`"
library_names = {}
```

JamesWrigley:
Alrighty, I did a second attempt in 891aff4. Now the auditor will check both `library_name` and the lib names in `library_names` for validity, and it'll only log a warning if `library_name` is not set and `library_names` doesn't contain entries for all the headers. I had to modify the `Audit` pass and `TranslationUnit` a bit to make that work since we don't store the passed headers in a `Context` (and I figured it made sense to store the file path with its `TranslationUnit`).

src/trans_unit.jl
```diff
@@ -6,6 +6,8 @@ Parse the given source file and the translation unit corresponding to that file.
 mutable struct TranslationUnit
     ptr::CXTranslationUnit
     idx::Index
+    source_filename::String
```
Gnimuc, line 9: this might break someone's custom generator which uses `TranslationUnit` exported by Clang.jl.

src/generator/audit.jl
```diff
@@ -1,20 +1,41 @@
...
-        catch err
+        if !_valid_lib_name(general_options["library_name"])
             general_options["library_name"] = "libxxx"
             @warn "failed to parse the `library_name` as a valid Julia expression, default library name: \":libxxx\" is being used instead."
         end
-    else
+    elseif !issubset(headers, keys(library_names))
         general_options["library_name"] = "libxxx"
-        @warn "default libname: \":libxxx\" is being used, did you forget to set `library_name` in the toml file?"
+        @warn "default libname: \":libxxx\" is being used because `library_names` is missing entries, did you forget to set `library_name` in the toml file?"
```
Gnimuc, line 28: this warning is a bit misleading. it sounds like `library_names` needs to be an exhaustive list, but actually not necessary to be.

to keep it simple and stupid, maybe it's a good idea to let users always define `library_name` to suppress the warning.

Gnimuc: to change the old warning to something like:

```
@warn "default libname: \":libxxx\" is being used, did you forget to set `library_name` in the toml file? It's safe to ignore this warning if you are using `library_names` as an exhaustive list."
```

JamesWrigley: Ok, I changed it in af62eff.

src/trans_unit.jl
JamesWrigley: I've reverted the change in af62eff, but how would this break a custom generator? :thinking:

Gnimuc: very unlikely, but possible if they've extended a `source_filename` field via `getproperty`...

ViralBShah:
Thanks folks!

PR #479, opened 2024-04-05
Fix unhandled non-field struct children

serenity4:
Unhandled non-field child nodes were sometimes present in `children(cursor)`, and were already handled in printing but not in codegen, which triggered an assertion error at https://github.com/JuliaInterop/Clang.jl/blob/30aaa5de6be51c105dc80139533d0db67263b762/src/generator/print.jl#L119

where `fields` built from the codegened expression had an extra field for a forward struct declaration.

Here is an exemple header where the error appeared:

```c
typedef struct _ExampleStruct {
  struct _ForwardA *a;
} ExampleStruct;

typedef struct _ForwardA {
  int a;
} ForwardA;
```

and codegen produced something like
```julia
struct _ExampleStruct
  _ForwardA::_ForwardA
  a::Ptr{_ForwardA}
end
```

Now we don't have that weird spurious `_ForwardA::_ForwardA` field.

It happened to me while wrapping a larger header (`winuser.h`) using this pattern, but I couldn't come up with a minimal reproducer. For some reason, `fields(Clang.getCursorType(cursor))` for `_ExampleStruct` was empty in that larger header and Clang.jl fell back to using `children` to extract fields, while in any example I have tried for a minimal reproducer `fields(Clang.getCursorType(cusor))` succeeded in returning the only field present as expected.

### DIFF
--- a/src/generator/codegen.jl
+++ b/src/generator/codegen.jl
@@ -493,13 +493,18 @@ function emit_constructor!(dag, node::ExprNode{<:StructLayout}, options)
     push!(node.exprs, func)
 end
 
+function extract_fields(cursor::CLCursor)
+    field_cursors = fields(getCursorType(cursor))
+    !isempty(field_cursors) && return field_cursors
+    # `chldren(cursor)` may also return forward declaration of struct type for example, so we filter these out.
+    filter(x->x isa CLFieldDecl, children(cursor))
+end
+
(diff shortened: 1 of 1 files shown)


Gnimuc:
`_emit_getproperty_ptr!` may also need to be patched.

Gnimuc:
Could be related to the upstream changes discussed in #472

Gnimuc:
let's wait for more bug reports.

PR #481, opened 2024-04-07
Update the changelog

JamesWrigley:
I left the release date empty for now since I'm not sure when v18 will be tagged, but the date should be fixed beforehand. I hope my summaries are correct :sweat_smile:


JamesWrigley:
I think the nightly failures are unrelated, perhaps in JSON3.jl.

PR #485, opened 2024-04-16
Add flag to control outputting of system code

### LINKED ISSUE #484: Unsure how to handle JLL dependencies
I'm working on generating bindings for a family of JLLs (all the AWS CRT ones). Some of them depend on each other. For example, `aws_c_io_jll` depends on `aws_c_common_jll`. In order to get Clang.jl to generate the bindings for `aws_c_io_jll`, I need to include both it and its dependent JLLs:
```julia
for target in JLLEnvs.JLL_ENV_TRIPLES
    options = load_options(joinpath(@__DIR__, "generator.toml"))
    options["general"]["output_file_path"] = joinpath(@__DIR__, "..", "lib", "$target.jl")

    header_dirs = []
    args = get_default_args(target)
    inc = JLLEnvs.get_pkg_include_dir(aws_c_common_jll, target)
    push!(args, "-I$inc")
    inc = JLLEnvs.get_pkg_include_dir(aws_c_io_jll, target)
    push!(args, "-I$inc")
    push!(header_dirs, inc)

    headers = String[]
    for header_dir in header_dirs
        for (root, dirs, files) in walkdir(header_dir)
            for file in files
                if endswith(file, ".h")
                    push!(headers, joinpath(root, file))
                end
            end
        end
    end
    unique!(headers)

    ctx = create_context(headers, args, options)
    build!(ctx)
end
```

This makes sense from the perspective of Clang.jl needing to fully resolve the DAG to emit valid code, but it comes with a problem: the bindings for `aws_c_common_jll` are included in the output. We're trying to create a composable set of bindings packages, so we'd like to find a way to output only the bindings for `aws_c_io_jll` and include the bindings for the dependencies via `using LibAwsCommon` in the generated code (another bindings package we have).

To my understanding, this is not a feature Clang.jl currently has. Please do correct me if I'm wrong there. I've explored Clang.jl's implementation a bit but it is not clear to me how I could implement this. Do you have any ideas on where I could start?

cc @quinnj

Octogonapus:
This PR adds a flag `output_system` (`true` by default for backwards compatibility) which controls whether code in system headers will be output.
Setting this flag to false allows the user to effectively avoid emitting code for dependencies by including those dependencies as system headers.

I have tested this with my MWE in #484 and it solves that issue.

Fixes #484.

### DIFF
--- a/src/generator/passes.jl
+++ b/src/generator/passes.jl
@@ -907,8 +907,10 @@ function (x::Audit)(dag::ExprDAG, options::Dict)
     return dag
 end
 
-function should_exclude_node(node, ignorelist, exclusivelist)
+function should_exclude_node(node, ignorelist, exclusivelist, isystem_ignorelist=[])
     str_node = string(node.id)
+    str_node ∈ isystem_ignorelist && return true
+
     for item ∈ ignorelist
         the_match = match(Regex(item), str_node)
         if the_match !== nothing  && the_match.match == str_node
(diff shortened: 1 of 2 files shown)


Gnimuc:
I changed the option name to `generate_isystem_symbols` and tweaked the code a little.

src/generator/passes.jl
```diff
@@ -943,6 +943,7 @@ function (x::FunctionPrinter)(dag::ExprDAG, options::Dict)
     show_info = get(log_options, "FunctionPrinter_log", x.show_info)
     ignorelist = get(general_options, "output_ignorelist", get(general_options, "printer_blacklist", []))
     exclusivelist = get(general_options, "output_exclusivelist", nothing)
+    get(general_options, "generate_isystem_symbols", true) && append!(ignorelist, string(x.id) for x in dag.sys)
```
Octogonapus, line 946: I believe this logic is inverted

Octogonapus:
After fixing the logic and trying it out, this new implementation is very slow because the ignorelist is quite large and running all of those regular expressions is slow. I would recommend to use the prior implementation because it is much faster.

Gnimuc:
@Octogonapus fixed again.

PR #488, opened 2024-04-25
Print explicit import for using CEnum

JoshuaLampert:
With the `print_using_CEnum` option equal `true`, an implicit import for `CEnum` is produced by printing `using CEnum`. However, then https://github.com/ericphanson/ExplicitImports.jl complains about the implicit import. It would be nice if `@cenum` (and probably also `CEnum`) would be imported explicitly. Since `@cenum` is the only exported name by CEnum.jl this approach should not break anything. Correct me if I am wrong.


Gnimuc:
Thanks!

PR #500, opened 2024-06-28
Add support for function-like macros in the auditors sanity checks

### LINKED ISSUE #499: `sanity check failed` thrown when generating bindings for `lexbor`
The error message requested filing an error here. This may, perhaps, be user-error rather than a problem with the package :)

Here's a reproducible example of the failure: https://github.com/MichaelHatherly/lexbor_clang_issues. See the CI failure in the following link: https://github.com/MichaelHatherly/lexbor_clang_issues/actions/runs/9664367322/job/26658816002#step:5:180, which also happens locally for me.

Perhaps the

```
┌ Warning: [CollectDependentSystemNode]: found symbols in the system headers: [:intptr_t]
└ @ Clang.Generators ~/.julia/packages/Clang/uJQ2L/src/generator/passes.jl:95
```

warning that prints prior to the error has something to do with it.

Potentially useful information:

- Upstream C library is https://github.com/lexbor/lexbor.
- JLL package is https://github.com/JuliaBinaryWrappers/lexbor_jll.jl
- Ygg build scripts is https://github.com/JuliaPackaging/Yggdrasil/blob/b1d2ca338a2e4004d792f80b6e94f14929b3f245/L/lexbor/build_tarballs.jl

JamesWrigley:
Fixes #499. The problem is that in [one of the headers](https://github.com/lexbor/lexbor/blob/master/source/lexbor/css/syntax/token.h) there are macros with the same names as struct types, e.g.:
```c
// Line 29
#define lxb_css_syntax_token_delim(token) ((lxb_css_syntax_token_delim_t *) (token))

// Line 139
typedef struct lxb_css_syntax_token_delim {
    lxb_css_syntax_token_base_t base;
    lxb_char_t                  character;
}
lxb_css_syntax_token_delim_t;
```

We already check for these kinds of conflicts between structs and functions, but not structs and macro-like functions so the auditor failed. I think this ought to be fixed upstream by changing the struct definitions to use untagged structs like so:
```c
typedef struct {
    lxb_css_syntax_token_base_t base;
    lxb_char_t                  character;
}
lxb_css_syntax_token_delim_t;
```

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -16,6 +16,8 @@ Changelog](https://keepachangelog.com).
 
 - The generator will now explicitly import the symbols from `CEnum` it uses to
   avoid implicit imports ([#488]).
+- Added support to the auditor for detecting structs and function-like macros of
+  the same name, which previously caused the generator to crash ([#500]).
 
 ## [v0.18.3] - 2024-04-23
 
--- a/src/generator/audit.jl
+++ b/src/generator/audit.jl
@@ -40,13 +40,15 @@ struct bar {
 typedef struct foo bar;
 typedef struct bar foo;
 ```
-* use the same name for structs and functions. For example,
+* use the same name for structs and functions/function macros. For example,
 ```c
 struct post {
     char m;
 };
 
-void post(post m);
+#define post(x) ((int)(x))
+// Or
+void post(struct post m);
 ```
 * two identifiers with same name are actually different things.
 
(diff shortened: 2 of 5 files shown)


Gnimuc:
LGTM.

Gnimuc:
cc @MichaelHatherly

MichaelHatherly:
Thanks @JamesWrigley.

PR #503, opened 2024-08-09
Generate bindings for Clang 18(Julia 1.12)

christiangnrd:
Master branch is now using LLVM 18 so add those bindings.


Gnimuc:
Thanks!

PR #506, opened 2024-08-20
Add compiler version selection support in `get_default_args`

Gnimuc:
also fix clang exec dirs

### DIFF
--- a/src/generator/context.jl
+++ b/src/generator/context.jl
@@ -245,7 +245,7 @@ function get_default_args(triple=get_triple(); is_cxx=false, version=GCC_MIN_VER
     if is_cxx
         env = get_default_env(triple; version, is_cxx)
         args = ["-isystem" * dir for dir in get_system_includes(env)]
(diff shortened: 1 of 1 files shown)


src/generator/context.jl
```diff
@@ -241,11 +241,15 @@ function get_triple()
...
 end
 
-function get_default_args(triple=get_triple(); is_cxx=false)
+function get_default_args(triple=get_triple(); is_cxx=false, version=GCC_MIN_VER)
     if is_cxx
-        env = get_default_env(triple; version=GCC_MIN_VER, is_cxx)
+        env = get_default_env(triple; version, is_cxx)
         args = ["-isystem" * dir for dir in get_system_includes(env)]
-        clang_inc = joinpath(LLVM_LIBDIR, "clang", string(Base.libllvm_version.major), "include")
+        @static if Base.VERSION <= v"1.10"
```
christiangnrd, line 248: Can this be removed already since earliest supported version is now 1.11?

PR #513, opened 2024-10-26
Add a warning about field_access_method_list not being portable

JamesWrigley:
I may take a stab at fixing this later, but for now we can at least warn the users.

### DIFF
--- a/gen/generator.toml
+++ b/gen/generator.toml
@@ -178,7 +178,11 @@ use_ccall_macro = true
 # if true, variadic functions are wrapped with `@ccall` macro. Otherwise variadic functions are ignored.
 wrap_variadic_function = false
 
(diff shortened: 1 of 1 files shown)


JamesWrigley:
The failure on x86 nightly does not spark joy :eyes: From looking at the testsets list I'm guessing it failed at the `File` test, but I don't see anything 32bit-specific that would fail. Could just be a spurious nightly failure given that the other x86 tests passed.

Gnimuc:
It would be better to add warnings in the codegen code so people can not ignore them. 

This is what we do to warn people about platform-specific definitions in system headers.

https://github.com/JuliaInterop/Clang.jl/blob/8ed2dd214fd89748b92a13b656b5f00b3fd71a18/src/generator/passes.jl#L95

PR #516, opened 2024-11-24
Widen integer literal to unsigned when required

### LINKED ISSUE #515: Integer literals incorrectly treated as `Clong`
I observed that macro definitions translated like the following:

```c
#define TEST 0x80000001L
```
to
```julia
const TEST = Clong(0x80000001)
```

fail with an `ERROR: InexactError: convert(Int32, 0x80000001)` because we are wrapping the literal with `Clong` (`:= Int32`). AFAIK C compilers will treat the literal as a `Culong` (`:= UInt32`), detecting that the literal is too large for a `Clong`. The solution would be to apply the same rule, and in this case, wrap with a `Culong`.

Any pointers as to where to perform the fix?

serenity4:
Fixes #515.

This include an additional fix with `txt_no_sfx = replace(txt_no_sfx, r"^0X" => "0x")` to also support integer constants formatted as `0X012345678`, which otherwise throw an error when executed.


src/generator/macro.jl
```diff
@@ -66,7 +66,11 @@ function c_non_float_literal_to_julia(sfx)
     elseif occursin("ul", sfx) || occursin("lu", sfx)
         return "Culong"
     elseif !unsigned && endswith(sfx, "l")
-        return "Clong"
+        if length(txt_no_sfx) == 10 && startswith(txt_no_sfx, r"0x[8-9a-fA-F]")
```
JamesWrigley, line 69: I think it might be simpler to try parsing the string as a `Clong` and falling back to `Culong` if that fails.

JamesWrigley (commented):
Would be nice to make a changelog entry too :)

src/generator/macro.jl
serenity4: True, it would also help with non-hexadecimal representations that may still represent large numbers.

serenity4:
Nightly failures seem unrelated, locally the fix handles hexadecimal and decimal literal integers correctly (using `Clong` by default and only wrap in `Culong` when the value doesn't fit).

JamesWrigley:
Sorry forgot to mention, could we add a test too? I would recommend adding a header file in https://github.com/JuliaInterop/Clang.jl/tree/master/test/include and then using it in https://github.com/JuliaInterop/Clang.jl/blob/master/test/generators.jl.

serenity4:
Sure, thanks for the pointer! Should be good now.

serenity4:
All good :)

JamesWrigley:
Unofficial LGTM!

Gnimuc:
Thanks！

PR #518, opened 2024-11-28
Improve performance for large `printer_blacklist` entries

serenity4:
Regular expressions were compiled on the fly for every entry in `printer_blacklist` every time a new node needed to see if its symbol was blacklisted.

This dramatically speeds up expression printing when there are more than a few entries in `printer_blacklist`, which prior to this change jumped from a total of ~20 seconds to several minutes when wrapping a large API (and after this change is back to ~20 seconds for this specific case).

### DIFF
--- a/src/generator/passes.jl
+++ b/src/generator/passes.jl
@@ -907,12 +907,12 @@ function (x::Audit)(dag::ExprDAG, options::Dict)
     return dag
 end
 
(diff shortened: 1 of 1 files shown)


Gnimuc:
re-triggered tests

JamesWrigley:
@serenity4 BTW I think the nightly tests will pass now if you rebase on master.

Gnimuc:
Thanks!

PR #520, opened 2024-12-07
Add char types to enum alias link

christiangnrd:
Feel free to close if I'm incorrect, but I think the "primitive enum values" should be the same in the `LinkEnumAlias` pass as the ones from cursor.jl?

https://github.com/JuliaInterop/Clang.jl/blob/572eddfbb30a30fdf2257ca7ec11a146380dbc6c/src/cursor.jl#L245-L255


Gnimuc:
LGTM

PR #519, opened 2024-12-07
Short and Elaborated Enums

christiangnrd:
Adds support for enums for shorts and aliased types. This is part of a series of PRs that will hopefully enable generating wrappers from ObjectiveC headers.

PR #524, opened 2024-12-17
Enable ObjectiveC framework enum and struct parsing

christiangnrd:
Builds on/requires #522.

This PR (along with #519 and #522) are the minimal changes required to allow Metal.jl to autogenerate the enums and structs for our frameworks. See https://github.com/JuliaGPU/Metal.jl/pull/492

I had to split up `create_context` because we use the same passes, but could not use the first part up to and including the header parsing.


Gnimuc:
This needs to be rebased on the latest master. Maybe just cherry-pick the commit.

Gnimuc:
done

PR #527, opened 2025-01-25
Objective-C property wrapping

christiangnrd:
Adds what I need to wrap Objective-C Interfaces and Protocols and their properties. No methods yet.

I did my best to split this up in reasonable commits to make review easier and includes some tests.

### DIFF
--- a/src/Clang.jl
+++ b/src/Clang.jl
@@ -33,6 +33,8 @@ foreach(names(@__MODULE__; all=true)) do s
         @eval export $s
     end
 end
+export CLVersion, CLPlatformAvailability
+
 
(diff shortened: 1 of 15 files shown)


Gnimuc:
I never used Objective-C, does it share the same compilation environment/flags with C? I mean do we need a new `ObjectiveCContext <: AbstractContext` dedicated to it.

christiangnrd:
To be honest I don't know how much is different. I was able to do most

Gnimuc:
Is this ready to be merged? I think the context stuff can be added later.

christiangnrd:
In that case it is

PR #528, opened 2025-02-13
support larger bit field

chengchingwen:
This PR reworks the bit field get/set property code to support bit field larger than 32 bits

### DIFF
--- a/src/generator/codegen.jl
+++ b/src/generator/codegen.jl
@@ -299,6 +299,7 @@ end
 function _emit_getproperty_ptr!(body, root_cursor, cursor, options)
     field_cursors = fields(getCursorType(cursor))
     field_cursors = isempty(field_cursors) ? children(cursor) : field_cursors
+    bitoffset = 0
     for field_cursor in field_cursors
         n = name(field_cursor)
         if isempty(n)
@@ -311,11 +312,13 @@ function _emit_getproperty_ptr!(body, root_cursor, cursor, options)
         offset = getOffsetOf(getCursorType(root_cursor), n)
         if isBitField(field_cursor)
             w = getFieldDeclBitWidth(field_cursor)
-            @assert w <= 32 # Bit fields should not be larger than int(32 bits)
-            d, r = divrem(offset, 32)
-            ex = :(f === $(QuoteNode(fsym)) && return (Ptr{$ty}(x + $(4d)), $r, $w))
+            d = (offset - bitoffset) ÷ 8
+            r = bitoffset
+            bitoffset += w
+            ex = :(f === $(QuoteNode(fsym)) && return (Ptr{$ty}(x + $d), $r, $w))
         else
             d = offset ÷ 8
+            bitoffset = 0
             ex = :(f === $(QuoteNode(fsym)) && return Ptr{$ty}(x + $d))
         end
         push!(body.args, ex)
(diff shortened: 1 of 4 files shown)


chengchingwen:
@christiangnrd Could you help review this PR?

christiangnrd:
I'm not familiar enough with this part of the codebase to review this. I gave it a try but it would take more bandwidth than I have at the moment.

src/generator/codegen.jl
```diff
@@ -311,9 +311,10 @@ function _emit_getproperty_ptr!(body, root_cursor, cursor, options)
         offset = getOffsetOf(getCursorType(root_cursor), n)
         if isBitField(field_cursor)
             w = getFieldDeclBitWidth(field_cursor)
-            @assert w <= 32 # Bit fields should not be larger than int(32 bits)
-            d, r = divrem(offset, 32)
-            ex = :(f === $(QuoteNode(fsym)) && return (Ptr{$ty}(x + $(4d)), $r, $w))
+            iszero(w) && continue
+            d, r = divrem(offset, 8)
+            @assert r + w <= 128 "Cannot handle bitfield larger than 128 bits"
```
Gnimuc, line 316: Is this documented in the standard?

chengchingwen: I don't think so. It's just super rare for a c/c++ code to have a primitive type that is larger than 128 bits, so I thought it would be reasonable to be left unhandled until we find a real use case.

Gnimuc: The test only has 64-bit test cases. Do you have a specific need for the 128-bit support?

chengchingwen: No, I only need 64-bit support

test/bitfield/bitfield.h
```diff
@@ -19,7 +19,27 @@ struct Mirror {
   unsigned int f;
 };
 
+struct LargeBitField {
+  double a;
+  short b;
+  unsigned long long c : 16;
```
Gnimuc, line 25: @chengchingwen It looks like you removed the logic for handling straddle. Does it work if you change this to 20?

`unsigned long long c : 16;`

Gnimuc: 20+48>64, so `d` overlaps another word.

chengchingwen: @Gnimuc The reason that the old logic has a branch for straddle is because they only perform 32-bit load on 32-bit-aligned "offset" thus a bit field would span across two 32-bit segments as long as the offset is smaller than multiples of 32 and adding the bit width exceeds multiples of 32. The new logic performs variable-sized load on 8-bit-aligned "offset" based on the required bit size.  

The bitfield "offset" obtained by `getOffsetOf` is not the start point of the whole bitfield but a point determined by the compiler that could be a middle point of that bitfield. Having 20 + 48 wouldn't necessarily cause `d` to overlap another "word" because of alignment and padding. Meanwhile, the "offset" obtained might just exceed the 20 bit part.

Gnimuc: @chengchingwen It looks like the old logic is for supporting AIX layout. 

https://github.com/llvm/llvm-project/blob/0defd832eb7c0618a67556e6fcbd32dd19e88b97/clang/lib/AST/RecordLayoutBuilder.cpp#L1460-L1528

I think we can add an option to let users select which layout they need. 

> The bitfield "offset" obtained by getOffsetOf is not the start point of the whole bitfield but a point determined by the compiler that could be a middle point of that bitfield. Having 20 + 48 wouldn't necessarily cause d to overlap another "word" because of alignment and padding. Meanwhile, the "offset" obtained might just exceed the 20 bit part.

As we use the value from Clang, could you give it check whether your current impl is compatible with Clang? https://github.com/llvm/llvm-project/blob/main/clang/lib/AST/RecordLayoutBuilder.cpp#L1534

chengchingwen:
temporary closed. I plan to update this PR based on clang's `LayoutBitField` and find a better way to test this feature recently.

PR #530, opened 2025-03-09
Handle typedef-ed unions nested in struct correctly

### LINKED ISSUE #529: union types represented as `NTuples` do not align correctly
Hello,

I have some wrappers generated by Clang, with some pretty complex data structures that involve many unions.
Here is a MWE:
```c
#include <stdio.h>
#include <stddef.h>

typedef union {
  int i;
  long l;
  long long l64;
  double d[2];
} BIGTYPE;

typedef struct {
  UCHAR type;
  UCHAR flags;
  UCHAR flags2;
  BIGTYPE value;
} C_STRUCT;

int main() {
	C_STRUCT c;
	c.type = 2;
	c.value = (BIGTYPE)255;

	printf("%s: %ld\n", "sizeof(var): ",sizeof(C_STRUCT));
	printf("%s: %ld\n", "sizeof(var.value): ",sizeof(BIGTYPE));

	printf("%s: %ld\n", "offsetof(var.flags): ",offsetof(C_STRUCT, flags));
	printf("%s: %ld\n", "offsetof(var.flags2): ",offsetof(C_STRUCT, flags2));
	printf("%s: %ld\n", "offsetof(var.value): ",offsetof(C_STRUCT, value));

	printf("actual value: %i", c.value.i);

	return 0;
}
```
compiling with `gcc` and running obtains:
```
sizeof(c): : 24
sizeof(c.value): : 16
offsetof(c.flags): : 1
offsetof(c.flags2): : 2
offsetof(c.value): : 8
actual value: 255
```

Clang automatically wraps the union type as a struct with a `NTuple`
```julia
struct BIGTYPE
    data::NTuple{16, UInt8}
end

struct C_STRUCT
    type::Cuchar
    flags::Cuchar
    flags2::Cuchar
    value::BIGTYPE
end
```
But the issue is that apparently `NTuple`s align with respect to its eltype, in this case it aligns as a `UInt8`, while C unions align w.r.t. the largest alignment of the types inside the union. Which in this case the union aligns as a `UInt128`.

This issue can be see easily with:
```julia-repl
julia> fieldoffset(C_STRUCT, 1)
0
julia> fieldoffset(C_STRUCT, 2)
1
julia> fieldoffset(C_STRUCT, 3)
2
julia> fieldoffset(C_STRUCT, 4)
3   # should be 8!
```

if instead we model the union as a `UInt128`
```julia
struct BIGTYPE
    data::UInt128
end
```
we get the correct behavior: `fieldoffset(C_STRUCT, 4) == 8`

But this got me thinking, why can't we use `primitive types` to model c unions instead?
I've tried with
```julia
primitive type BIGTYPE 16 * 8 end
```
as a slot in replacement of the `NTuple` version, and worked quite well both in terms of alignment and the overloaded `getproperty` accessor.

What gotchas am i not seeing?

Gnimuc:
fix #529

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.7'
+          - '1.11'
           - '1'
           - 'nightly'
         os:
--- a/src/generator/passes.jl
+++ b/src/generator/passes.jl
@@ -679,7 +679,13 @@ function (x::CodegenPreprocessing)(dag::ExprDAG, options::Dict)
                 # if any dependent node is a `RecordLayouts` node, mark this node as `StructLayout`
                 for j in n.adj
                     dn = dag.nodes[j]
-                    dn.type isa RecordLayouts || continue
+                    should_mark = dn.type isa RecordLayouts
+                    # resolve typedefs that are not marked correctly
+                    if is_typedef_elaborated(dn)
+                        dn_cur = getTypeDeclaration(getTypedefDeclUnderlyingType(dn.cursor))
+                        should_mark = dn_cur isa CLUnionDecl
+                    end
+                    should_mark || continue
                     # ignore pointer fields
                     field_cursors = fields(getCursorType(n.cursor))
                     field_cursors = isempty(field_cursors) ? children(n.cursor) : field_cursors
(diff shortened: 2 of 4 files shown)


Gnimuc:
I think the failure on 1.7 is related to https://github.com/JuliaInterop/Clang.jl/commit/1eb99ce07275bc2d481e7302c2eefcc9d64b5840 

don't have time to dive into this. Can we drop the old Julia version support?

PR #533, opened 2025-03-10
Work around #532

### LINKED ISSUE #532: Test suite segfault when run with more than 1 thread.
While trying to run the tests on master, macOS tests were failing (different reason), and the linux tests were throwing a segmentation fault basically right away. After lots of playing around,  I tried running with `JULIA_NUM_THREADS=1` and the testsuite passed.

This should either be fixed or a warning should print when tests are run with more than 1 thread.

Reproduced on linux x86_64 (Intel 12900) and linux aarch64 (RPi 5), does not fail on macOS.

christiangnrd:
This seems to be an issue with running concurrent tests when the spinner is enabled. This PR deactivates the spinner. This closes #532 and makes the test output somewhat cleaner.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -7,4 +7,4 @@ using REPL
 
 include("ClangTests.jl")
 
-retest(Clang, ClangTests; stats=true)
+retest(Clang, ClangTests; stats=true, spin=false)


christiangnrd:
> Unofficial LGTM! I'll try looking into this later, ReTest.jl is supposed to work with multiple threads.

Sounds good! I think this should stay even when ReTest gets fixed since the Clang.jl test suite is quite verbose and the spinner makes the output quite hard to read through.

I also just opened https://github.com/JuliaTesting/ReTest.jl/issues/61.

Gnimuc:
@JamesWrigley feel free to merge if it get fixed.

JamesWrigley (approved):
Roger that, I'll merge this now so as to fix the tests.

PR #542, opened 2025-04-08
throw an error instead of falling back to `getfield` for `Ptr` types

ghyatzo:
Hello,

The generated `Base.getproperty(x::Ptr{T}, f::Symbol)` emits a `getfield(x,f)` if all other cases are exhausted.
But `getfield` is not really defined for `Ptr` (which makes sense, since it doesn't have any field) and static analysis tools like `JET` generate a lot of noise about it.

What about throwing an exception instead? It shouldn't be breaking since we'd be throwing an exception anyway, and provides a better error message and makes analysis tools happy.


Gnimuc:
This fallback pattern is used ubiquitously in Julia Base.

ghyatzo:
Yes, It's a common pattern. And I think it makes absolute sense when overloading the `getproperty` of objects that do have fields.

But when used to give an object virtual fields I think that same pattern is a bit misplaced.

Anyway, this is minor, it would just be to play nicer with static analyzers. I'll leave this here if you change mind!

Issue #535, opened 2025-03-20
struct definitions with unions broken in some cases?


thomvet:
```
typedef struct {
    UA_ExtensionObjectEncoding encoding;
    union {
        struct {
            UA_NodeId typeId;   /* The nodeid of the datatype */
            UA_ByteString body; /* The bytestring of the encoded data */
        } encoded;
        struct {
            const UA_DataType *type;
            void *data;
        } decoded;
    } content;
} UA_ExtensionObject;
```

Results in:
```
struct var"union (unnamed at headers\\open62541\\types.h:890:5)"
    data::NTuple{40, UInt8}
end

function Base.getproperty(x::Ptr{var"union (unnamed at headers\\open62541\\types.h:890:5)"}, f::Symbol)
    f === :encoded && return Ptr{var"struct (unnamed at headers\\open62541\\types.h:891:9)"}(x + 0)
    f === :decoded && return Ptr{var"struct (unnamed at headers\\open62541\\types.h:895:9)"}(x + 0)
    return getfield(x, f)
end
```

I am certain that parsing this in earlier versions has worked, but it seems broken now. Currently on Julia 1.11.4 and Clang.jl 0.18.3.

thomvet:
Just tested this. It works with Julia 1.10.9 and Clang.jl 0.17.8.

PR #544, opened 2025-06-23
Fix `getproperty` and `propertynames` in certain circumstances

christiangnrd:
I believe 9610bf5bf96fb7267374086f6a7373846d35680b may have broken this.


JamesWrigley:
RIP CI lmao :sweat_smile: I'll try to have a look at the failures on 1.12+ later. In which circumstances does this fix `getproperty`/`propertynames`?

JamesWrigley:
Oh, so this is a Clang change that we have to adapt to?

christiangnrd:
I believe so. I think this essentially fixes an extension of https://github.com/JuliaInterop/Clang.jl/issues/535

PR #545, opened 2025-07-31
Update changelog for v0.19.0

JamesWrigley:
I think it's time we make a release to support Julia 1.12 since that's coming out soon-ish.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,6 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v2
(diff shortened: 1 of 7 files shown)


JamesWrigley:
Still need to figure out the segfaults, but in any case I think we should merge #544 first.

JamesWrigley:
@Gnimuc, any objections if I merge #544 and then implement the above?

Gnimuc:
> @Gnimuc, any objections if I merge #544 and then implement the above?

LGTM. Feel free to merge.

Gnimuc (approved):

JamesWrigley:
Nightly is failing because Julia is on LLVM 20 now. But we can add support for that later, I think it's better to get a release out soon with support for 1.12.

PR #552, opened 2025-09-01
Add mapping for intptr_t to JuliaCintptr_t

wsmoses:
cc @giordano @gbaraldi @jumerckx


wsmoses:
@JamesWrigley can this get a review/merge/patch?

JamesWrigley (changes requested):
Thanks, looks good :+1: Could you add a changelog entry to https://github.com/JuliaInterop/Clang.jl/blob/master/docs/src/changelog.md?

PR #563, opened 2026-03-03
Switch to Clang_unified_jll.

maleadt:
Clang_jll has a couple of issues:
- each Julia version requires a different Clang_jll version, generating unnecesary churn when switching Julia versions.
- Clang_jll only works correctly if the compat entries in General are correct, which they often aren't, e.g., right now Clang_jll v21 gets installed on Julia 1.14 despite it only using LLVM v20.

Fix this by switching to a unified JLL that uses platform augmentations to select a compatible artifact based on whatever LLVM version was detected.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,17 +1,17 @@
 name = "Clang"
 uuid = "40e3b903-d033-50b4-a0cc-940c62c95e31"
-version = "0.19.2"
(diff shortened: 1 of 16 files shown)


JamesWrigley (approved):
Sweet, that's handy. LGTM!

PR #575, opened 2026-06-21
ObjectiveC: Don't add 'immutable' flag.

maleadt:
With https://github.com/JuliaInterop/ObjectiveC.jl/pull/80, the default behavior is preferable.

cc @christiangnrd

### DIFF
--- a/src/generator/codegen.jl
+++ b/src/generator/codegen.jl
@@ -927,7 +927,7 @@ end
     end
 
     # @objcwrapper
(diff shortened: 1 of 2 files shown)


JamesWrigley (changes requested):
I think the CI failures are unrelated so feel free to ignore them, I'll have a look later. This looks like a breaking change so I think it should have a changelog entry? Feel free to also bump the version number if you want to tag a release.

christiangnrd:
Seems reasonable to me

PR #577, opened 2026-07-30
Fix `Clang.LLVM_INCLUDE`, and disable LLVM's signal handlers

JamesWrigley:
See the individual commits for details. I also enabled multiple threads in CI, which probably would have caught the signal handler issue before.

### DIFF
--- a/src/Clang.jl
+++ b/src/Clang.jl
@@ -41,6 +41,8 @@ include("string.jl")
 include("file.jl")
 export CLFile, name, unique_id, get_filename
 
+include("signals.jl")
+
 include("index.jl")
 export Index
 
@@ -91,5 +93,14 @@ export LLVM_VERSION, LLVM_LIBDIR, LLVM_INCLUDE, CLANG_INCLUDE
 
 include("generator/Generators.jl")
 using .Generators
+
+function __init__()
+    try
+        preregister_llvm_signal_handlers()
+    catch ex
+        @warn "Could not pre-register LLVM's signal handlers, this can \
+               cause segfaults when running Julia with multiple threads." exception=(ex, catch_backtrace())
+    end
+end
 
 end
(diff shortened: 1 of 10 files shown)


src/Clang.jl
```diff
@@ -84,12 +86,21 @@ end
...
 const CLANG_INCLUDE = LLVM_INCLUDE
 
 export LLVM_VERSION, LLVM_LIBDIR, LLVM_INCLUDE, CLANG_INCLUDE
 
 include("generator/Generators.jl")
 using .Generators
 
+function __init__()
+    try
+        preregister_llvm_signal_handlers()
```
vchuravy, line 99: I am confused, we shouldn't be calling InitLLVM on a library path?

vchuravy: E.g. in the long time Julia has used LLVM we havent had to deal with this, so I think we should figure out why we are registering handlers in the libclang case when libLLVM doesnt

JamesWrigley: Is `InitLLVM()` called by `clang_enableStackTraces()`? I don't think Clang.jl explicitly calls it. The reason I noticed it is because I now run Julia with multiple threads by default, but I had Claude check other versions and this also happened on 1.11 and 1.12 when Julia is started with multiple threads.

Gnimuc: It looks like those handlers are installed here:

https://github.com/llvm/llvm-project/blob/4e0d78f974499e728de1ee1acbdce1e6491a610c/clang/tools/libclang/CIndex.cpp#L4097-L4098

https://github.com/llvm/llvm-project/blob/4e0d78f974499e728de1ee1acbdce1e6491a610c/llvm/lib/Support/CrashRecoveryContext.cpp#L150

controlled by an environment variable `LIBCLANG_DISABLE_CRASH_RECOVERY`.

JamesWrigley: I don't think it's that, we already disable those handlers: https://github.com/JuliaInterop/Clang.jl/blob/81ab6e05ae21cd53dbbc5149f7e8f7d883df3fb5/src/index.jl#L20

JamesWrigley:
Yggdrasil now has cmake v4.4 so I reverted the change to use the system cmake in 4a81293.

src/Clang.jl
vchuravy: Is that too late? We have this issue with other libraries, but we have sofar avoided something this complicated to work around it

